PR #119, opened 2014-12-24
(WebappPlugin) speed up service dev

### LINKED ISSUE #118: (WebappPlugin) add task `reStartService`
Currently, issuing `reStart` in a project that enables the `WebappPlugin` is slow because it always runs the `npm:build` task. This is especially painful when you are only working on the API (service) portion of the project and are not making any UI changes at all. It would be great to have a `reStart`-esque task that does not run the `npm:build` task.

This issue was brought up in https://github.com/allenai/s2-online/issues/20

markschaake:
Fixes #118 

I thought it made more sense to revert `reStart` to no longer depend on `npm:build` in favor of adding a new task `reStartWebapp` which runs `npm:build` and then `reStart`. This gives the developer the option to either 

1. build and run just the service with `reStart`, or
1. build and run both the service and webapp frontend with `reStartWebapp`

### DIFF
--- a/src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
@@ -23,6 +23,11 @@ object WebappPlugin extends AutoPlugin {
       val logNodeEnvironment = TaskKey[Unit](
         "logNodeEnvironment",
         "Logs the NodeJs build environment to SBT console")
+
+      val reStartWebapp = TaskKey[Unit](
+        "reStartWebapp",
+        "Restarts the service without running npm:build"
+      )
     }
   }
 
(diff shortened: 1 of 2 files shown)


schmmd:
Cool, mind if Sam has a look though?

src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
```diff
@@ -23,6 +23,11 @@ object WebappPlugin extends AutoPlugin {
       val logNodeEnvironment = TaskKey[Unit](
         "logNodeEnvironment",
         "Logs the NodeJs build environment to SBT console")
+
+      val reStartWebapp = TaskKey[Unit](
+        "reStartWebapp",
+        "Restarts the service without running npm:build"
```
codeviking: Shouldn't this be something like: `"Restarts the service after rebuilding the front-end."`

src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
```diff
@@ -34,15 +39,18 @@ object WebappPlugin extends AutoPlugin {
     log.info(s"[webapp] NODE_ENV = '$env'")
   }
 
+  val reStartWebappTask = WebappKeys.reStartWebapp := {
+    (NodeKeys.build in Npm).value
+    Revolver.reStart.value
```
codeviking: This differs from the line below which was removed, but I'm assuming works as expected -- just making sure this is correct.

codeviking:
Just a tweak to wording, then this looks g2g.

codeviking:
Also, quick question.  Without really knowing how Revolver works, I assume `reStart` is built in.  Previously we were "extending" restart to depend upon the npm build, but now we just create a new task which runs the npm build and then the default `reStart` task included with revolver out of the door?

markschaake:
@codeviking yes that is correct - we are reverting to the default `reStart` task and supplying a new `reStartWebapp` task that does what we previously had extended `reStart` to do.

src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
markschaake: yep, thanks for catching that

src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
markschaake: Yes, this works.

markschaake:
Released version 2015.01.05-0

PR #120, opened 2015-01-02
Use env variable for AWS PEM file.

schmmd:


src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -220,7 +218,10 @@ object DeployPlugin extends AutoPlugin {
     // Command to pass to rsync's "rsh" flag, and to use as the base of our ssh
     // operations.
     val sshCommand = {
-      val sshKeyfile = configMap("deploy.user.ssh_keyfile")
+      val pemEnvVar = "AWS_PEM_FILE"
+      val sshKeyfile = sys.env.get(pemEnvVar).getOrElse {
```
jkinkead: I think the idea behind the `validateAsMap` function is that you can assume that this exists when you reach here, and don't need to re-validate.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -220,7 +219,8 @@ object DeployPlugin extends AutoPlugin {
     // Command to pass to rsync's "rsh" flag, and to use as the base of our ssh
     // operations.
     val sshCommand = {
-      val sshKeyfile = configMap("deploy.user.ssh_keyfile")
+      val pemEnvVar = "AWS_PEM_FILE"
+      val sshKeyfile = sys.env.get(pemEnvVar).get
```
jkinkead: Just use `apply`?

jkinkead:
LGTM! Feel free to merge, addressing the comment if you like.

PR #123, opened 2015-01-16
Set default scala version to 2.11.

schmmd:


markschaake:
LGTM

build.sbt
```diff
@@ -17,6 +17,7 @@ scalacOptions := Seq(
   "-language:_",
   "-Xlog-reflective-calls")
 
+// SBT requires 2.10 for now (1/15/15).
```
jkinkead: So, this sets the default version within the core dependencies plugin, but not for the plugins library itself? That makes sense.

jkinkead:
Non-sarcastic question! If the answer is "yes", LGTM! If it's not, I'd be confused, and would like to understand better what's going on.

build.sbt
schmmd: Yes.  The plugins library needs to compile for 2.10.  I get confused by this every time--I keep trying to bump it up to 2.11 but that's impossible.

jkinkead: Sweet. Merging.

PR #124, opened 2015-01-22
Update various versions

dirkgr:
This updates various versions of common dependencies. @markschaake, if you are busy, is there anyone else who maintains this plugin? Otherwise, do you mind taking a look?

### DIFF
--- a/src/main/scala/org/allenai/plugins/CoreDependencies.scala
+++ b/src/main/scala/org/allenai/plugins/CoreDependencies.scala
@@ -10,7 +10,7 @@ trait CoreDependencies {
   val defaultScalaVersion = "2.11.5"
 
   object Logging {
-    val slf4jVersion = "1.7.7"
+    val slf4jVersion = "1.7.10"
     val logbackVersion = "1.1.2"
     // The logging API to use. This should be the only logging dependency of any API artifact
     // (anything that's going to be depended on outside of this SBT project).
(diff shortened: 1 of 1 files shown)


dirkgr:
@jkinkead, if Mark is busy, can you take a look at this? I'm getting a bit deep into `publishLocal` hell with the rat's tail of changes that depend on this.

jkinkead:
All the changes look reasonable. Did you want me to test this locally or anything before an LGTM?

Is this driven by any particular requirement, like the switch to scala 2.11?

dirkgr:
Yes, this whole thing is about switching to 2.11. We don't have to update everything to do this, but I find that I can get rid of a ton of dependency overrides by doing this.

markschaake:
LGTM too!

src/main/scala/org/allenai/plugins/CoreDependencies.scala
```diff
@@ -79,7 +79,7 @@ trait CoreDependencies {
   val sprayCaching = sprayModule("caching")
 
   // Spray json (separate from Spray toolkit)
-  val sprayJson = "io.spray" %% "spray-json" % "1.2.6"
+  val sprayJson = "io.spray" %% "spray-json" % "1.3.1"
```
markschaake: This will be a headache as there are breaking API changes in 1.3. I think we should deal with it though since it provides 10x performance improvement.

dirkgr: The only change I've found is that `JsArray` now takes a `Vector` rather than a `List`. If it's for the sake of performance, sure, I guess, but that does seem a little specific.

Also, curse github for not showing this comment until after I hit the merge button.

dirkgr: Also, since when are API-breaking changes allowed in minor version changes? *shakes fist*

PR #128, opened 2015-02-15
Use the new scalariform sbt-plugin in the CoreSettingsPlugin.

jkinkead:
I'm a little worried that this still seems to pull in the old scalariform version in as a dependency; I suspect (as noted in `build.sbt`) that the exclude isn't working for some reason.

I think this should be compatible with the existing jar in the github hook package, but I'll republish in a bit to make sure.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,12 +1,7 @@
 import bintray.{ Keys => BintrayKeys }
 import bintray.{ Plugin => BintrayPlugin }
 
-libraryDependencies ++= Seq("com.typesafe" % "config" % "1.2.0",
-  // This seems to be required in order to pick up the newer scalariform version; either the exclude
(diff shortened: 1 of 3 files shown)


jkinkead:
Figured it out; I've run into this before. I need to include the scala version in the artifact name. :)

schmmd:
Changes look good.  I'll try this out on common-scala before merging.

schmmd:
Should this fix idempotency?  It didn't on the `extraction` repository but it did pretty well.

Could we make scalariform remove trailing whitespace and make sure there is a newline at the end of the file?

jkinkead:
It should fix the biggest idempotency issue (the one with comments). If you have an example file that takes > 1 iteration to format, that would be cool. I think I've fixed at least one bug around method signatures not working the first time, but I believe there are more.

PR #131, opened 2015-02-25
Don't addLoggingDependencies by default.

schmmd:


schmmd:
This burned @cristipp recently. @jkinkead meant it as a solution for Aristo, not a general solution to all logging issues. @dirkgr claims it doesn't work in some projects.

We can still manually `addLoggingDependencies` for now if needed.  Longer term we would optimistically come up with a better solution.

jkinkead:
For documentation purposes:

The function `addLoggingDependencies` needs to be called after *all* libraries are added to the current path or else it doesn't work - and furthermore, calling it a second time completely breaks things.

This should go in the subproject's `build.sbt` in the future, after all other dependencies are added:
```
addLoggingDependencies(libraryDependencies)
```

PR #133, opened 2015-03-12
Reorganize release versioning.

schmmd:


src/main/scala/org/allenai/plugins/ReleasePlugin.scala
```diff
@@ -14,32 +14,51 @@ object ReleasePlugin extends AutoPlugin {
...
+  object SemanticVersion {
+    def nextVersion(version: String) = {
+      sbtrelease.Version(version).
+        map(_.bump(Version.Bump.Next).asSnapshot.string).
+        getOrElse(sbtrelease.versionFormatError)
+    }
+    def releaseVersion(version: String) = {
+      sbtrelease.Version(version).
+        map(_.withoutQualifier.string).
+        getOrElse(sbtrelease.versionFormatError)
```
markschaake: Instead of objects, I think it would be better to provide ready-to-use `Seq[Def.Setting[_]]` here. Otherwise you force the build author to know to override two settings. Something like:

```scala
object autoImport {
  object Release {
    val dateVersioning: Seq[Def.Setting[_]] = Seq(
      releaseVersion := { version => DateVersion.releaseVersion(version) },
      nextVersion := { version => DateVersion.nextVersion(version) }
    )

    val semanticVersioning: Seq[Def.Setting[_]] = Seq(
      releaseVersion := { version => SemanticVersion.releaseVersion(version) },
      nextVersion := { version => SemanticVersion.nextVersion(version) }
    )
  }
}

override def projectSettings: Seq[Def.Setting[_]] = autoImport.Release.dateVersioning
```

Then in `build.sbt` you'd just do this:

```scala
enablePlugins(LibraryPlugin)
Release.semanticVersioning
```

schmmd:
I like having testable functions.  Here's a take 2 that provides an easy way to mix them in.

schmmd:
It probably would be nice, but it's too much overhead for me to learn right now ;-P

PR #134, opened 2015-04-02
Run npm prune prior to npm install to prevent dependency conflicts with ...

markschaake:
...no-longer-used libs

This was motivated by an observed failed built on Shippable where the error was caused by dependency conflicts between currently used lib and a no-longer-used lib. Seems like an npm bug, but this should solve it for us.


markschaake:
Here is a build that failed for this reason:

https://app.shippable.com/builds/551d8f9e23bf360c00588302

The error is that `browserify-shim` is not compatible with `browserify`. However, `browserify-shim` is not longer a dependency for the project. Pruning should fix this.

codeviking:
lgtm

PR #137, opened 2015-04-25
Relax conflict manager; add IntegrationTest config by default

markschaake:
This PR started as just an upgrade of Akka to 2.3.10, and then I decided to throw some more stuff in :)

- Upgrade akka version to 2.3.10
- Extend the `akkaModule` method to take a version override if you don't want the default
- Removed `ConflictManager.strict` from core settings. I thought we had decided to do this a while back, but could be wrong.
- Added `IntegrationTest` config by default to all projects (via `CoreSettings` autoplugin).

@schmmd and / or @jkinkead review?
@dirkgr any concerns with the conflict manager being non-strict (take latest) and our plans to move to semantic versioning?

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version = 0.13.7
+sbt.version = 0.13.8
--- a/src/main/scala/org/allenai/plugins/CoreDependencies.scala
+++ b/src/main/scala/org/allenai/plugins/CoreDependencies.scala
@@ -69,8 +69,16 @@ trait CoreDependencies {
   val typesafeConfig = "com.typesafe" % "config" % "1.2.1"
 
   // Akka
-  val akkaVersion = "2.3.9"
-  def akkaModule(id: String): ModuleID = "com.typesafe.akka" %% s"akka-$id" % akkaVersion
+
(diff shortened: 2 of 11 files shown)


schmmd:
> Removed ConflictManager.strict from core settings. I thought we had decided to do this a while back, but could be wrong.

We didn't exactly decide this, but we were leaning that direction.  I think this is a good change.

schmmd:
Looks great to me.  I plan to publish with semantic versions to BinTray soon (starting with common, maybe even today).  I don't think we need to change anything to enable this--am I wrong?

jkinkead:
w.r.t. the conflict manager, I thought there was some setting @dirkgr updated that overrode this or wasn't compatible. I think the eviction policy or something?

dirkgr:
Non-strict conflict manager is great. In fact, many of the projects I touch are already switched over, and they could now have that setting removed.

markschaake:
> don't think we need to change anything to enable this--am I wrong?

@schmmd  I don't think so... I think you're good to go

schmmd:
@markschaake you are assigned ;-)

markschaake:
Merged, will now publish

Issue #138, opened 2015-04-25
Upgrade DeployPlugin dependency on sbt-native-packger to 1.0.0


markschaake:
SBT native packager is now at 1.0.0 and is a major (breaking) release. It is now implemented with AutoPlugins, which should make it easier to work with.

http://www.scala-sbt.org/sbt-native-packager/

PR #139, opened 2015-04-25
Updgrade native packager

markschaake:
This is needed to finalize the consolidation of s2-online into s2-offline. I'm going to merge and release without review.

@schmmd @jkinkead FYI


markschaake:
I just realized I also snuck in an upgrade to Akka... this was accidental but shouldn't be a problem

markschaake:
Resolves #138

PR #141, opened 2015-04-29
Don't create the Universal start / stop scripts when staging for deploy.

jkinkead:
This is a confusing feature of the packager, especially because these scripts will silently disappear if you have multiple main methods configured.

@afader - I remember you being confused by the random scripts showing up in `bin`.


markschaake:
LGTM

PR #142, opened 2015-04-30
Publish to Bintray

schmmd:


schmmd:
@markschaake I would appreciate a look-over from you too.  I'm pretty convinced this will work for common.  I'm a little worried about what defaults I'm causing other builds to have.

schmmd:
Should I remove `bintray.Plugin.bintrayPublishSettings` so downstream applications specify this instead?  Is it OK to leave the bintray plugin in the allenai plugins?

markschaake:
Sorry, just noticed this... taking a look now

src/main/scala/org/allenai/plugins/ReleasePlugin.scala
```diff
@@ -62,11 +63,43 @@ object ReleasePlugin extends AutoPlugin {
...
-      releaseVersion := { version => SemanticVersion.computeReleaseVersion(version) },
-      nextVersion := { version => SemanticVersion.computeNextVersion(version) }
+      ReleaseKeys.releaseVersion := { version => SemanticVersion.computeReleaseVersion(version) },
+      ReleaseKeys.nextVersion := { version => SemanticVersion.computeNextVersion(version) }
     )
   }
 
-  override lazy val projectSettings: Seq[Def.Setting[_]] =
-    WrappedReleasePlugin.releaseSettings ++ DateVersion.settings
+  val checkBranchIsNotMaster = { st: State =>
```
jkinkead: A type on this `(State => State)` would be really helpful.

src/main/scala/org/allenai/plugins/ReleasePlugin.scala
```diff
@@ -62,11 +63,43 @@ object ReleasePlugin extends AutoPlugin {
...
+    val vcs = Project.extract(st).get(ReleaseKeys.versionControlSystem).getOrElse {
+      sys.error("Aborting release. Working directory is not a repository of a recognized VCS.")
+    }
+
+    if (vcs.currentBranch == "master") {
+      sys.error("Current branch is master.  At AI2, releases are done from another branch and " +
```
jkinkead: If you were writing this to be more open-source / user friendly, you could make the deploy branch be a setting key.

jkinkead:
A couple of small comments. I'm not sure what adding the bintray settings you're adding do, so I don't know if they'd be harmful at all.

src/main/scala/org/allenai/plugins/ReleasePlugin.scala
schmmd: This specifies which branch CANNOT be used for releases.  Setting is a good idea--I'll see how easy that is for me.

markschaake:
I also don't know what adding the bintray settings will do. We could just see what happens :)

LGTM

PR #143, opened 2015-04-30
Make blacklist configurable.

schmmd:


src/main/scala/org/allenai/plugins/ReleasePlugin.scala
```diff
@@ -12,9 +12,13 @@ import java.text.SimpleDateFormat
 import java.util.Date
 
 object ReleasePlugin extends AutoPlugin {
-
   override def requires: Plugins = plugins.JvmPlugin
 
+  val blacklistedBranches = SettingKey[Seq[String]](
```
jkinkead: Couldn't you make this a `Set` from the getgo?

src/main/scala/org/allenai/plugins/ReleasePlugin.scala
```diff
@@ -68,15 +72,18 @@ object ReleasePlugin extends AutoPlugin {
...
     val vcs = Project.extract(st).get(ReleaseKeys.versionControlSystem).getOrElse {
       sys.error("Aborting release. Working directory is not a repository of a recognized VCS.")
     }
 
-    if (vcs.currentBranch == "master") {
-      sys.error("Current branch is master.  At AI2, releases are done from another branch and " +
-        "then merged into master via pull request.  Shippable, our continuous build system does " +
-        "the actual publishing of the artifacts.")
+    val branch = vcs.currentBranch
+    println(branch)
```
jkinkead: Probably don't want the printlns in.

jkinkead:
A couple of comments, but otherwise LGTM!

PR #144, opened 2015-05-07
Update the deploy + node plugins.

jkinkead:
This is mostly to avoid the super-janky run-sbt-subprocess-from-within-sbt that was going on.

There are a few subtle functional changes, but I expect no one will notice them. I plan on sending out release notes with this, just in case.

Please feel free to bug me in-person or on github about anything you don't understand. I'm happy to share the sbt ~~burden~~ knowledge.

Changes:
* project name is taken from sbt settings, not deploy config
* run-class.sh is generated before the `stage` command now, instead of before compile
* deploy cleans the local `stage` directory only, since we should assume the rest of the build system works
* dev & deploy node build environment set with sbt setting


src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -1,90 +1,82 @@
...
-  * be issued to restart the server.
+  * This will replace _completely_ the contents of the remote {bin,conf,lib,public} directories with
+  * the contents from the build. All other directories and files (such as pid and log files) will be
+  * left intact.  After the new binaries are pushed, "${deploy.startup_script} restart" will be
+  * issued to restart the server.
   *
-  * There is a special ~/.deployrc config file that is merged into any deploy
-  * target configs as an override.
+  * There is a special ~/.deployrc config file that is merged into any deploy target configs as an
+  * override.
```
codeviking: So, do settings in `~/.deployrc` get merged into those put in `conf/deploy.conf`?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -1,90 +1,82 @@
...
+  * Argument overrides have precedence over .deployrc overrides.
   */
 object DeployPlugin extends AutoPlugin {
 
-  override def requires: Plugins = plugins.JvmPlugin && JavaAppPackaging
+  // JavaAppPackaging gives us the `stage` command, and NodeJsPlugin lets us deploy webapps.
+  override def requires: Plugins = plugins.JvmPlugin && JavaAppPackaging && NodeJsPlugin
+
+  /** Static usage string. */
+  val Usage = "Usage: deploy <overrides> [deploy target]";
```
codeviking: `<overrides`> should be optional right? `[<overrides?]`?

src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
```diff
@@ -113,26 +92,56 @@ object NodeJsPlugin extends AutoPlugin {
...
       case t: Throwable => state
     }
   }
 
+  /** @returns the node build environment variables, with the given NODE_ENV and NODE_BUILD_DIR set
+    */
+  def getEnvironment(nodeEnv: String, buildDir: File): Map[String, String] = {
+    Map(
+      "NODE_ENV" -> nodeEnv,
+      "NODE_API_HOST" -> "/api",
```
codeviking: Seems like this should be configurable too, just in case someone decides to use a different API root.   Though I also hope to eliminate this property in the front-end build system, so maybe we leave it as is for now.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -144,7 +140,25 @@ object DeployPlugin extends AutoPlugin {
...
-    Seq(file)
+    file
+  }
+
+  lazy val npmBuildTask = Def.taskDyn {
+    if ((nodeEnv in thisProject).value == "dev") {
+      Def.task {
+        NodeJsPlugin.execBuild(
+          (NodeKeys.nodeProjectDir in Npm).value,
+          NodeJsPlugin.getEnvironment("dev", (NodeKeys.nodeProjectTarget in Npm).value)
```
codeviking: I assume the second param to execBuild is the environment variables? (I'll check below just noting my assumption)

src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
```diff
@@ -14,37 +15,17 @@ import sbt.Keys._
...
-    // Set NODE_ENV to the deploy target (e.g. 'prod', 'staging', etc.)
-    NodeKeys.buildEnvironment in Npm := deployEnvironment.value.getOrElse("sbt-dev"),
-    // Print the node environment on stage
     // Force npm:build when using sbt-revolver re-start to ensure UI is built
     Revolver.reStart <<= Revolver.reStart.dependsOn(NodeKeys.build in Npm),
-    mappings in Universal <++= (NodeKeys.nodeProjectTarget in Npm) map MappingsHelper.directory,
-    mappings in Universal <<= (mappings in Universal).dependsOn(NodeKeys.build in Npm)
+    // Build the node project on stage (for deploys).
+    UniversalPlugin.autoImport.stage <<=
+      UniversalPlugin.autoImport.stage.dependsOn(DeployPlugin.npmBuildTask),
```
codeviking: :+1:

test-projects/test-deploy/build.sbt
```diff
@@ -1,11 +1,5 @@
...
-
-printEnv := {
-  val s = streams.value
-  val env = deployEnvironment.value
-  s.log.info(s"Environment: $env")
-}
-
-compile in Compile <<= (compile in Compile).dependsOn(printEnv)
+// Tester project for services with a node build.
+val nodeapp = project.in(file("nodeapp")).enablePlugins(WebappPlugin)
```
codeviking: The name `nodeapp` is a little confusing as it's not actually server-side js (I suppose it is to build), but it's meant to be a webapp...anyway clearly not a showstopper...

codeviking:
Looks good, some very minor questions / comments -- feel free to merge when you feel it's ready.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
jkinkead: Yeah, although the reason for this was ssh keys, which have since moved into the AWS_PEM_FILE environment variable - and which probably won't be needed in the mid-term future anyway.

This is something I'd like to drop, but I'm trying not to make this PR a "cleanup everything" monolith.

test-projects/test-deploy/build.sbt
jkinkead: This was purely because I found `webapp/webapp` unappealing to look at. I can switch back (there's not really a good reason for the name).

src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
jkinkead: I actually have no idea what this is, I just kept what was already there. :)

I feel like there might be a few places the node plugin could be made better. OK with leaving as-is, then?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
jkinkead: Yep. It mimics the parameters to the private `exec` method in the node plugin.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
jkinkead: Fixing!

PR #145, opened 2015-05-07
Move clean & test dependencies to webapp plugin.

jkinkead:
Make this more consistent (and play more nicely with the deploy plugin).


codeviking:
:+1:

PR #146, opened 2015-05-08
Remove unused config parameter.

jkinkead:
This is never read.


src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -410,7 +410,7 @@ object DeployPlugin extends AutoPlugin {
       }
     }
     // Coerce optional keys into strings.
-    val optionalKeys = Seq("project.subdirectory", "project.version")
```
codeviking: :+1:

PR #147, opened 2015-05-08
Gulpiness!

jkinkead:
So gulpy.

This does what we discussed during lunch: `npm install; npm watch` run before the first `re-start`; subsequent `re-start` calls do nothing.

`re-stop` kills the background watch.

### DIFF
--- a/src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
@@ -7,9 +7,25 @@ import sbt.inc.Analysis
 // imports standard command parsing functionality
 import complete.DefaultParsers._
 
+import java.lang.{ Runtime => JavaRuntime }
+
+import scala.collection.mutable
+
 /** Plugin that provides an 'npm' command and tasks for test, clean, and build.
   */
 object NodeJsPlugin extends AutoPlugin {
+  private val watches = mutable.Map.empty[ProjectRef, Process]
+  // Set up a shutdown hook to kill any running processes when we exit.
+  JavaRuntime.getRuntime.addShutdownHook(
+    new Thread(new Runnable {
+      override def run(): Unit = {
+        watches.synchronized {
+          watches.values.foreach { _.destroy() }
+          watches.clear()
+        }
+      }
+    })
+  )
 
   object autoImport {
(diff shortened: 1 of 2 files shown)


src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
```diff
@@ -21,6 +37,12 @@ object NodeJsPlugin extends AutoPlugin {
         "Execution `npm install` in the Node application directory to install dependencies"
       )
 
+      // Note that this isn't called 'watch' because there's another SBT key with this name and a
+      // different value.
+      val nwatch = taskKey[Unit]("Executes `npm watch` in the Node directory")
```
codeviking: I think it technicall runs "npm run watch" but it's not crucial to update that.

codeviking:
This is awesome!  One small remark which is trivial -- merging!

src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
```diff
@@ -23,8 +23,10 @@ object WebappPlugin extends AutoPlugin {
   override lazy val projectSettings: Seq[Setting[_]] = Seq(
     // Expect the node project in a "webapp" subdirectory.
     NodeKeys.nodeProjectDir in Npm := (baseDirectory in thisProject).value / "webapp",
-    // Force npm:build when using sbt-revolver re-start to ensure UI is built
-    Revolver.reStart <<= Revolver.reStart.dependsOn(NodeKeys.build in Npm),
+    // Run "npm watch" when we run a re-start.
+    Revolver.reStart <<= Revolver.reStart.dependsOn(NodeKeys.nwatch in Npm),
```
markschaake: What happens when I'm running `> ~reStart` ? Will `NodeKeys.nwatch` get called each time? Does `reStart` trigger `reStop`?

jkinkead: `re-start` has stopping baked into the task - it doesn't call out to `re-stop`: https://github.com/spray/sbt-revolver/blob/master/src/main/scala/spray/revolver/Actions.scala#L26-L30

As discussed offline, `nwatch` *does* get called each time, but it's smart enough not to spin up extra external processes.

PR #151, opened 2015-05-12
Adds DatabasePlugin

### LINKED ISSUE #149: Add a non-archetype DatabasePlugin
We now have several projects that depend on both slick and flyway for database integration and database migration, respectively. This means we have duplicate boilerplate across project build files. The solution: create a plugin of course!

The `DatabasePlugin` should not be an archetype, but instead a "mixin" plugin. Usage would be something like this:

```scala
// in build.sbt
enablePlugins(WebappPlugin, DatabasePlugin)
```

## DatabasePlugin Spec

The `DatabasePlugin` should:

* add slick as a library dependency (which version - 2.1 or 3.0? I think 3.0 even though it was just released)
* add flyway as a build dependency
* add flyway settings to the build
* apply conventions (such as migration directory)
* provide a task to generate slick tables from a connected db (we have done this, for example, here: https://github.com/allenai/scholar/blob/e0e21bb594e93324e41547773778bdd17b6cf695/data/sources/arxiv/slick-codegen.sbt)

As a reference for necessary flyway settings, you can look here: https://github.com/allenai/scholar/blob/e0e21bb594e93324e41547773778bdd17b6cf695/data/crawler/flyway.sbt

markschaake:
Resolves #149

This adds a non-archetype auto plugin for easing set up of database and migration.

@dirkgr want this one?

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -1,0 +1,60 @@
+DatabasePlugin
+====================
+
+**Required by**: None
+
+**Requires**: CoreSettings
+
+The `DatabasePlugin` provides [Slick](http://slick.typesafe.com/) dependencies and [Flyway](http://flywaydb.org/) database migration.
+
+## Usage
+
+### build.sbt set up
+
+```scala
+// in build.sbt
+enablePlugins(DatabasePlugin)
+
+// you need to provide a JDBC driver for flyway migrations
+libraryDependencies += org.allenai.plugins.CoreDependencies.postgresDriver
+
+// Migrations:
+// Specify a database URL for performing migrations. Only local dev
+// URL should be checked in to source control. Migrating production databases
+// should be a carefule one-time event that is preceded by a backup.
+flywayUrl := "jdbc:postgresql://localhost:5432/databaseplugin"
+flywayUser := "ai2dev"
+flywayPassword := "ai2dev"
+```
+
+### Migrations
+
+The `DatabasePlugin` provides the `sqlMigrationsInit` task that will create a directory
+where you will put your SQL migration scripts. It will also create an initial migration
+file with comments to help you understand the conventions.
+
+1. Initialize db-migrations directory:
+
+  ```shell
+  $ sbt
+  > sqlMigrationsInit
+  [info] Generating initial SQL migration file in /Users/markschaake/ai2/git/ai2-sbt-plugins/test-projects/test-database/db-migrations/V1__initial-schema.sql
+  [success]
+
+2. If the database does not already exist, make sure to create it (and the user you specify in build.sbt)
+
+3. Run flywayMigrate:
+
+  ```shell
+  $ sbt
+  > flywayMigrate
+  ```
+
+Note: you can always "clean" your local database, but this will remove all tables and data.
+
+  ```shell
+  $ sbt
+  > ;flywayClean;flywayMigrate
+  ```
+
+TODO: add a code generation task to generate Scala source code from table schema.
--- a/docs/database.md
+++ b/docs/database.md
@@ -5,7 +5,7 @@ DatabasePlugin
 
 **Requires**: CoreSettings
 
-The `DatabasePlugin` provides [Slick](http://slick.typesafe.com/) dependencies and [Flyway](http://flywaydb.org/) database migration.
+The `DatabasePlugin` provides [Slick](http://slick.typesafe.com/) dependencies and [Flyway](http://flywaydb.org/) for database migration.
 
 ## Usage
 
--- a/src/main/scala/org/allenai/plugins/CoreDependencies.scala
+++ b/src/main/scala/org/allenai/plugins/CoreDependencies.scala
@@ -95,6 +95,20 @@ trait CoreDependencies {
   // Spray json (separate from Spray toolkit)
   val sprayJson = "io.spray" %% "spray-json" % "1.3.1"
 
+  // Slick for database integration
+  // TODO consider upgrading to slick 3.0. All of our existing projects use
+  // slick 2.1.0, but we should considering migrating to 3.0.
+  val defaultSlickVersion = "2.1.0"
+
+  /** Generates a slick module dependency */
+  def slickModule(id: String, version: String = defaultSlickVersion): ModuleID =
+    "com.typesafe.slick" %% id % version
+
+  val slick = slickModule("slick")
+  val slickCodegen = slickModule("slick-codegen")
+
+  val postgresDriver = "org.postgresql" % "postgresql" % "9.4-1201-jdbc41"
+
   val loggingDependencyOverrides = Set(
     Logging.slf4jApi,
     Logging.logbackCore,
--- a/test-projects/test-database/build.sbt
+++ b/test-projects/test-database/build.sbt
@@ -1,0 +1,15 @@
+enablePlugins(DatabasePlugin)
+
+// You have to provide the database driver that corresponds to the JDBC URL
+libraryDependencies += org.allenai.plugins.CoreDependencies.postgresDriver
+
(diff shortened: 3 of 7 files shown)


src/main/scala/org/allenai/plugins/CoreDependencies.scala
```diff
@@ -95,6 +95,20 @@ trait CoreDependencies {
   // Spray json (separate from Spray toolkit)
   val sprayJson = "io.spray" %% "spray-json" % "1.3.1"
 
+  // Slick for database integration
+  // TODO consider upgrading to slick 3.0. All of our existing projects use
+  // slick 2.1.0, but we should considering migrating to 3.0.
```
jkinkead: @ekolve was having issues with exception propagation when testing 3.0 yesterday.

ekolve: After talking with marks and looking here: http://slick.typesafe.com/doc/3.0.0/api/index.html#slick.jdbc.JdbcBackend$DatabaseDef it should be possible to recover{} from a Failure since a Future is returned from the run() method.

docs/database.md
```diff
@@ -0,0 +1,60 @@
+DatabasePlugin
+====================
+
+**Required by**: None
+
+**Requires**: CoreSettings
+
+The `DatabasePlugin` provides [Slick](http://slick.typesafe.com/) dependencies and [Flyway](http://flywaydb.org/) database migration.
```
codeviking: Flyway *for* database migration?

codeviking:
:+1: Looks good to me.  Maybe one very small wording change, but otherwise g2g.

codeviking:
One question question -- could / does flyway handle non SQL migrations?  I'm thinking about the EF `mongodb` instance specifically, and whether we could start using `.js` scripts and thus mangle the json using javascript which (for me at least) tends to be easier.

codeviking:
Also, @markschaake want to add a persistent-webapp template to allenai/templates?  Would be sweet to have an example with this baked in!

test-projects/test-database/build.sbt
```diff
@@ -0,0 +1,15 @@
+enablePlugins(DatabasePlugin)
+
+// You have to provide the database driver that corresponds to the JDBC URL
+libraryDependencies += org.allenai.plugins.CoreDependencies.postgresDriver
+
+// In order migratinos to work, you need to have postgres running
+// at the following URL, and have a database already created
+// named 'databaseplugin'
+flywayUrl := "jdbc:postgresql://localhost:5432/databaseplugin"
```
dirkgr: Should this be an integration test then? I don't think this will work in Shippable?

dirkgr: On second thought, this could be made to run in shippable if you create a `databaseplugin` database. I think Shippable gives you a running postgres.

markschaake: I wouldn't think we'd want to run database migrations within shippable. But, we could if we wanted to test that our migrations were valid I suppose.

markschaake:
> Also, @markschaake want to add a persistent-webapp template to allenai/templates?

Good call, I'll put an issue into the templates project (don't have time to tackle that right now)

dirkgr:
The changes themselves look fine, if you're committed to slick and Flyway.

My main concern is that I don't think Flyway is a good idea, and I'm having doubts about slick.

I have a "system" for versioning database schema versions that I've used many times in the past, and I don't find it onerous at all. Basically, I have a `settings` table in every database, wherein I have a version number that versions the schema. When I connect to the database, I check whether the version is the expected version, and while it isn't, I run an update function that I pick out of a table of such functions. My update functions can be SQL or Scala, which means I can do migrations that can't easily be expressed in SQL. It also keeps schema code in one place, whereas with Flywheel we'll have the schema twice, once in Flywheel's migration code, and once in slick's table definitions. The method requires no extra dependencies, no documentation to read, no build steps, and it's very little code. I don't know what Flyway can offer that makes it worth to pay this cost.

I have similar concerns about slick, but weaker ones. I might need to feel the JDBC pain once more before I admit that slick is a good idea, but all the slick code in OKC I've written was problematic. A good abstraction is supposed to free me from thinking about a particular detail, but slick doesn't do that. I have a SQL query in my head, and I have to translate it into slick, so now I have to think about SQL as well as slick. The only thing it frees me from is thinking about format conversions between Scala's classes and how they're stored in SQL, but even that only starts working once you also depend on `slick-pg` and go through their [game of traits](https://github.com/allenai/okcorpus/blob/master/src/main/scala/org/allenai/dictionary/persistence/OkcPostgresDriver.scala). Slick is using some amazing Scala sorcery to do what it does, and as with most Scala sorcery, it means that their error messages are completely meaningless when you get something wrong. That's a high price to pay for automatic marshaling. If I were to build persistence for OKC again, I wouldn't use slick.

dirkgr:
Hmm, you already merged it. Now it's too late to even say this: Add `slick-pg` to the dependencies. It's necessary to use many great postgres features.

markschaake:
> Add slick-pg to the dependencies. It's necessary to use many great postgres features.

I want the plugin to be database vendor agnostic (could also be used with H2, for example). I only added the Postgres driver as a CoreDependency because we use Postgres in several projects.

markschaake:
> My main concern is that I don't think Flyway is a good idea, and I'm having doubts about slick.

We are using both Flyway and Slick in several projects. This plugin will reduce build boilerplate. As for opinions about what tools are the best for the long run, we can have that conversation in perhaps a future engineering meeting. I think a PR is not the appropriate place to have this discussion.

jkinkead:
I think PRs are a great place to have discussions over technology choices *especially* if it's like this PR where we're adding a blessed set of libraries to a core plugin. If there isn't consensus on which technologies to use, they shouldn't be in our blessed set in the archetype plugins.

This was merged already, unfortunately.

Since this was your PR, would it make sense for you to lead the discussion on why these two technologies were chosen, and Dirk can present alternatives?

markschaake:
> I think PRs are a great place to have discussions over technology choices especially if it's like this PR where we're adding a blessed set of libraries to a core plugin. If there isn't consensus on which technologies to use, they shouldn't be in our blessed set in the archetype plugins.

I agree PRs are a great place to discuss technology choices. The fact is we already use both Slick and Flyway in several projects, and I would consider this PR providing a service to those projects by reducing build boilerplate, not making a technology choice.

Also, I did not make this an _archetype plugin_, it is an opt-in mixin plugin (which is why it isn't in the archetypes package). The sole purpose of adding it here is to reduce build boilerplate in several projects that are currently using these technologies. I don't see the harm in having other mixin plugins available if it solves the boilerplate problem - regardless of whether they are already 'blessed' or still in 'trial' phase.

I think

dirkgr:
I'm happy to discuss alternatives on any platform.

My alternative to Flywheel is the manual process with the settings table I outlined above.

I don't have a good alternative to Slick, except for using JDBC directly. Yesterday, @ekolve alerted me to [Anorm](https://www.playframework.com/documentation/2.1.1/ScalaAnorm). I haven't used it, but it looks like it lets you write normal SQL, but does the SQL/Scala data type conversions for you. I would want to try this next time I write a new database thing.

jkinkead:
Sorry, I said "archetype plugin" but I really meant "core plugin". I don't think there's much distinction here, since having it in one of the central plugins is implicitly blessing the technology there.

Mostly I don't want to adopt tech here that people have serious reservations about, and it's good to acknowledge when we kinda settle on some tech choices. I'd much more prefer to spend 5 minutes discussing something like this at the eng meetings before sending a PR, and see if it warrants further discussion or not.

I'm fine leaving this as-is - but I didn't want any concerns Dirk has to be swept under the rug.

PR #154, opened 2015-05-13
Allow build to have the JavaAppPackager bash script written

markschaake:
@jkinkead what do you think? This will unblock me upgrading plugins in S2.


jkinkead:
Looks great!

I'd add a comment somewhere noting that this is a great script for run-once processes, but not the first choice for daemons.

markschaake:
@jkinkead comments and docs added. I'll merge away

PR #160, opened 2015-05-26
Make npm log level configurable; default to 'warn'

markschaake:
@schmmd this makes `npm` log level explicit and configurable. Before, we were using the `--quiet` flag which is an alias to `--loglevel warn`, which didn't actually do anything because default loglevel for npm is `warn`. Now, we can increase verbosity via:

```scala
// build.sbt

npmLogLevel in Npm := NpmLogLevel.Info

// or:
npmLogLevel in Npm := NpmLogLevel.Verbose

// or even:
npmLogLevel in Npm := NpmLogLevel.Silly
```

### DIFF
--- a/src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
@@ -27,6 +27,9 @@ object NodeJsPlugin extends AutoPlugin {
     })
   )
 
+  // from npm docs https://docs.npmjs.com/misc/config#loglevel
+  val ValidLogLevels = Set("silent", "error", "warn", "http", "info", "verbose", "silly")
+
   object autoImport {
     val Npm = ConfigKey("npm")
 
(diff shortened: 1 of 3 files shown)


src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
```diff
@@ -213,8 +236,8 @@ object NodeJsPlugin extends AutoPlugin {
   /** Execute `cmd` with `npm` setting `root` as the working directory.
     * @throws Exception if the process returns an non-zero exit code
     */
-  private def exec(cmd: String, root: File, env: Map[String, String]): Unit = {
-    fork(cmd, root, env).exitValue() match {
+  private def exec(cmd: String, root: File, env: Map[String, String], npmLogLevel: String): Unit = {
+    fork(s"$cmd --loglevel $npmLogLevel", root, env).exitValue() match {
```
jkinkead: Validate the log level? I wouldn't want to cause a weird build failure because of a misconfigured setting . . .

jkinkead:
Looks great! One comment.

And yes, I know this is assigned to Michael - but it was too exciting. :)

src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
markschaake: good call, will do

schmmd:
Cool--re-assigning to Jesse since he has more to say about this ;-)

markschaake:
@jkinkead ping! I've added the log level validation. Are we good to go?

markschaake:
@jkinkead hold off on this. I want to change how I'm doing validation.

markschaake:
@jkinkead back to you. I've handled validation on build load via `NpmLogLevel` type instead of at build time.

jkinkead:
Awesome, LGTM!

Merging.

schmmd:
Mark S, ping me when we have a release?

markschaake:
@schmmd Released version 2015.05.26-0

https://github.com/allenai/sbt-plugins/releases/tag/v2015.05.26-0

PR #162, opened 2015-05-29
(CoreDependencies) Upgrade to latest versions of spray and spray-json

markschaake:
The upgrades are bugfix releases. There are important fixes to `spray.http.Uri` that will benefit the S2 Crawler.

### DIFF
--- a/src/main/scala/org/allenai/plugins/CoreDependencies.scala
+++ b/src/main/scala/org/allenai/plugins/CoreDependencies.scala
@@ -84,7 +84,7 @@ trait CoreDependencies {
   val akkaTestkit = akkaModule("testkit")
 
   // Spray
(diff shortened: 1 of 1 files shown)


jkinkead:
LGTM!

What kind of fixes? I always found the `Uri` API a little disappointing.

markschaake:
@jkinkead sorry I meant to link the change log. Here it is: http://spray.io/project-info/changelog/#version-1-1-3-2015-03-24

markschaake:
and for Spray JSON: https://github.com/spray/spray-json/blob/master/CHANGELOG

jkinkead:
Dang, I was hoping for some methods to add parameters to URIs, or add path components, or go from the `spray.routing` Path to Uri and back again . . .

PR #163, opened 2015-05-29
(CoreSettings) make src/it depend on src/test classpath by default

markschaake:
While writing tests, we often create some helpers for setting up test cases. These helpers may be useful in integration tests as well as unit tests. Previously, integration tests did not have a classpath dependency on unit test classes. This PR makes it so! Now we can have test helpers in our unit tests that are also available to our integration tests.


jkinkead:
Awesome, that's a really sensible default. Merging!

PR #167, opened 2015-06-19
Make the StylePlugin more sane.

### LINKED ISSUE #166: Probable race condition in style check plugin
The style check step of our continuous build sometimes fails with a `SAXParseException`. This is thrown when reading the XML configuration file fed into scalastyle.

It's likely that there's a race condition when running parallel style checks - either running `styleCheck` at the same time as `test:styleCheck` and `it:styleCheck`, or else running parallel style checks across subprojects.

jkinkead:
I'm hoping this will fix #166 , since it totally skips the step where it writes out an XML file & reads it back in.

### DIFF
--- a/src/main/scala/org/allenai/plugins/StylePlugin.scala
+++ b/src/main/scala/org/allenai/plugins/StylePlugin.scala
@@ -3,28 +3,27 @@ package org.allenai.plugins
 import sbt._
 import sbt.Keys._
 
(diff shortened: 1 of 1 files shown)


markschaake:
LGTM. Merge when ready and release?

jkinkead:
Done! And thanks. :)

PR #168, opened 2015-06-23
Update release plugins to the latest version.

jkinkead:
Remove dependency on ai2 plugins.

Note that I had to rename the release plugin, because it conflicted with the auto-imported `ReleasePlugin`.

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "2015.06.19-1-SNAPSHOT"
+version in ThisBuild := "0.0.1"
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ AI2 SBT Plugins
 The AI2 SBT plugins are intended to minimize build boilerplate accross projects. It is recommended to only enable "Archetype" plugins, which currently include:
 
 - `CliPlugin` - for command line applications using the [scopt](https://github.com/scopt/scopt) library.
-- `LibraryPlugin` - for libraries to be released / published using our `ReleasePlugin`
+- `LibraryPlugin` - for libraries to be released / published using our `Ai2ReleasePlugin`
 - `WebServicePlugin` - for web service applications built on spray, akka, and spray-json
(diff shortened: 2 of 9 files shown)


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "2015.06.19-1-SNAPSHOT"
\ No newline at end of file
+version in ThisBuild := "0.0.1"
```
markschaake: Are you changing the versioning for our plugins to semantic in this PR?

markschaake:
LGTM other than what looks like a switch to semantic versioning. Were you intending to do that in this PR as well? If so, is this something that's been talked about?

version.sbt
jkinkead: Yes, because the date versioning was provided by the plugin I removed - and since Michael switched us to semantic versioning by default on other projects, it seemed reasonable to do here as well.

jkinkead:
Comment added about semantic versioning!

I'm going to hold off on merging until I can shepherd the release through & send out notes - this will require changing publication settings on any project using Bintray.

version.sbt
markschaake: > Yes, because the date versioning was provided by the plugin I removed

Ah right. SGTM, though an initial version of 0.0.1 "feels" slightly wrong. It is all very arbitrary, but when I see 0.0.1 I think this is a pre-alpha-release. I don't think we've talked about an AI2 policy for semantic versioning (i.e. when we consider something worthy of 1.0 status). Another idea is to change our group id + artifact id at this time to `"org.allenai" % "plugins" % VERSION` to enforce a clean separation from our date-based versioned artifacts.

Thoughts?

BTW, feel free to merge with 0.0.1 version. I don't have strong feelings but this seems like a good opportunity to discuss AI2 semantic versioning policy.

jkinkead: My plan was to do a verification deploy at 0.0.1, then bump up to 1.0.0 once I confirm the setup works. :)

markschaake: SGTM!

jkinkead:
Merging & proceeding with upgrade!

markschaake:
woo hoo 1.0! :+1:

PR #170, opened 2015-07-03
Add a function to get settings to disable bintray.

jkinkead:
Autoplugins confuse me. Apparently, just having the bintray plugin on the classpath installs it - which makes anything trying to publish to nexus kinda break (see [this semaphore log](https://semaphoreci.com/allenai/extraction/branches/master/builds/7) for an example). I say "kinda" break, because publication itself actually succeeds - sbt just reports a failure. That's probably a bug in the bintray plugin.

Anyway, this just adds a function you can call to remove the problematic Bintray tasks.


markschaake:
LGTM

PR #171, opened 2015-07-06
CacheKey calculation and Injection

ryanai3:
added tasks to calculate cacheKeys and inject them.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,7 @@ addSbtPlugin(
 addSbtPlugin("com.github.jkinkead" % "sbt-scalariform" % "0.1.6")
 
 // Dependency graph visualiztion in SBT console
-addSbtPlugin("com.gilt" % "sbt-dependency-graph-sugar" % "0.7.4")
+addSbtPlugin("com.gilt" % "sbt-dependency-graph-sugar" % "0.7.5")
 
 // Wrapped by WebServicePlugin and WebappPlugin
 addSbtPlugin("io.spray" % "sbt-revolver" % "0.7.2")
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,7 @@ addSbtPlugin(
 addSbtPlugin("com.github.jkinkead" % "sbt-scalariform" % "0.1.6")
 
 // Dependency graph visualiztion in SBT console
-addSbtPlugin("com.gilt" % "sbt-dependency-graph-sugar" % "0.7.5-1")
+addSbtPlugin("com.gilt" % "sbt-dependency-graph-sugar" % "0.7.4444")
 
 // Wrapped by WebServicePlugin and WebappPlugin
 addSbtPlugin("io.spray" % "sbt-revolver" % "0.7.2")
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,7 @@ addSbtPlugin(
 addSbtPlugin("com.github.jkinkead" % "sbt-scalariform" % "0.1.6")
 
 // Dependency graph visualiztion in SBT console
-addSbtPlugin("com.gilt" % "sbt-dependency-graph-sugar" % "0.7.5")
+addSbtPlugin("com.gilt" % "sbt-dependency-graph-sugar" % "0.7.5-1")
 
 // Wrapped by WebServicePlugin and WebappPlugin
 addSbtPlugin("io.spray" % "sbt-revolver" % "0.7.2")
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -124,7 +124,10 @@ object DeployPlugin extends AutoPlugin {
     IO.delete((UniversalPlugin.autoImport.stagingDirectory in Universal).value)
   }
 
-  val stageAndCacheKeyTask = stageAndCacheKey := VersionInjectorPlugin.injectCacheKeyTask
+  val stageAndCacheKeyTask = stageAndCacheKey := {
+    VersionInjectorPlugin.injectCacheKeyTask
+    println("") // so that we can be a Task[Unit]
+  }
 
   lazy val npmBuildTask = Def.taskDyn {
     if ((nodeEnv in thisProject).value == "dev") {
@@ -262,6 +265,7 @@ object DeployPlugin extends AutoPlugin {
     // Create the required run-class.sh script before staging.
     UniversalPlugin.autoImport.stage <<=
       UniversalPlugin.autoImport.stage.dependsOn(CoreSettingsPlugin.autoImport.generateRunClass),
+    stageAndCacheKeyTask,
 
     // Add root run script.
     mappings in Universal += {
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -64,6 +64,8 @@ object DeployPlugin extends AutoPlugin {
       "Cleans the staging directory. This is not done by default by the universal packager."
     )
 
+    val stageAndCacheKey = taskKey[Unit]("Checksums non-local jars in stage, & uses git commits to store cacheKey")
+
     /** The reason this is a Setting instead of just including * is that including * in the rsync
       * command causes files created on the server side (like log files and .pid files) to be
       * deleted when the rsync runs, which we don't want to happen.
@@ -122,6 +124,8 @@ object DeployPlugin extends AutoPlugin {
     IO.delete((UniversalPlugin.autoImport.stagingDirectory in Universal).value)
   }
 
+  val stageAndCacheKeyTask = stageAndCacheKey := VersionInjectorPlugin.injectCacheKeyTask
+
   lazy val npmBuildTask = Def.taskDyn {
     if ((nodeEnv in thisProject).value == "dev") {
       Def.task {
@@ -145,6 +149,7 @@ object DeployPlugin extends AutoPlugin {
   lazy val deployTask = deploy := {
     // Dependencies
     gitRepoClean.value
+    stageAndCacheKey.value
 
     val log = DeployLogger(streams.value.log)
 
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -124,10 +124,7 @@ object DeployPlugin extends AutoPlugin {
     IO.delete((UniversalPlugin.autoImport.stagingDirectory in Universal).value)
   }
 
-  val stageAndCacheKeyTask = stageAndCacheKey := {
-    VersionInjectorPlugin.injectCacheKeyTask
-    println("") // so that we can be a Task[Unit]
-  }
+  val stageAndCacheKeyTask = stageAndCacheKey := VersionInjectorPlugin.injectCacheKeyTask
 
   lazy val npmBuildTask = Def.taskDyn {
     if ((nodeEnv in thisProject).value == "dev") {
@@ -152,7 +149,7 @@ object DeployPlugin extends AutoPlugin {
   lazy val deployTask = deploy := {
     // Dependencies
     gitRepoClean.value
-    stageAndCacheKey.value
+    VersionInjectorPlugin.injectCacheKeyTask.value
 
     val log = DeployLogger(streams.value.log)
 
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -64,8 +64,6 @@ object DeployPlugin extends AutoPlugin {
       "Cleans the staging directory. This is not done by default by the universal packager."
     )
 
-    val stageAndCacheKey = taskKey[Unit]("Checksums non-local jars in stage, & uses git commits to store cacheKey")
-
     /** The reason this is a Setting instead of just including * is that including * in the rsync
       * command causes files created on the server side (like log files and .pid files) to be
       * deleted when the rsync runs, which we don't want to happen.
@@ -123,8 +121,6 @@ object DeployPlugin extends AutoPlugin {
   val cleanStageTask = cleanStage := {
     IO.delete((UniversalPlugin.autoImport.stagingDirectory in Universal).value)
   }
-
-  val stageAndCacheKeyTask = stageAndCacheKey := VersionInjectorPlugin.injectCacheKeyTask
 
   lazy val npmBuildTask = Def.taskDyn {
     if ((nodeEnv in thisProject).value == "dev") {
@@ -262,7 +258,6 @@ object DeployPlugin extends AutoPlugin {
     // Create the required run-class.sh script before staging.
     UniversalPlugin.autoImport.stage <<=
       UniversalPlugin.autoImport.stage.dependsOn(CoreSettingsPlugin.autoImport.generateRunClass),
-    stageAndCacheKeyTask,
 
     // Add root run script.
     mappings in Universal += {
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -149,7 +149,7 @@ object DeployPlugin extends AutoPlugin {
   lazy val deployTask = deploy := {
     // Dependencies
     gitRepoClean.value
-    VersionInjectorPlugin.injectCacheKeyTask.value
+    VersionInjectorPlugin.autoImport.injectCacheKey.value
 
     val log = DeployLogger(streams.value.log)
 
--- a/src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
@@ -14,7 +14,7 @@ object VersionInjectorPlugin extends AutoPlugin {
   object autoImport {
     val injectVersion = TaskKey[Seq[File]](
       "injectVersion",
-      "Generates the artifact.conf, git.conf, and cacheKey.conf version files"
+      "Generates the artifact.conf and git.conf version files"
     )
     val injectArtifact = TaskKey[File](
       "injectArtifact",
@@ -44,17 +44,13 @@ object VersionInjectorPlugin extends AutoPlugin {
       "gitDescribe",
       "The description of the current git commit"
     )
-    val gitMRC = TaskKey[String](
-      "gitMRC",
+    val gitLocalSha1 = TaskKey[String](
+      "gitLocalSha1",
       "Most recent commit in src"
     )
     val cacheKey = TaskKey[String](
       "cacheKey",
       "cacheKey of project"
-    )
-    val mTask = TaskKey[Seq[String]](
-      "fucksbt",
-      "it'sterrible"
     )
   }
 
@@ -70,7 +66,7 @@ object VersionInjectorPlugin extends AutoPlugin {
     injectCacheKeyTask,
     injectGitTask,
     gitDescribeTask,
-    gitMRCTask,
+    gitLocalSha1Task,
     gitCommitDateTask,
     gitSha1Task,
     gitRemotesTask,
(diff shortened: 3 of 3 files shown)


ryanai3:
@Jessie - please review

ryanai3:
@jkinkead please review

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -122,6 +124,11 @@ object DeployPlugin extends AutoPlugin {
     IO.delete((UniversalPlugin.autoImport.stagingDirectory in Universal).value)
   }
 
+  val stageAndCacheKeyTask = stageAndCacheKey := {
+    VersionInjectorPlugin.injectCacheKeyTask
+    println("") // so that we can be a Task[Unit]
```
jkinkead: This should be inferred without needing a special block.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -145,6 +152,7 @@ object DeployPlugin extends AutoPlugin {
   lazy val deployTask = deploy := {
     // Dependencies
     gitRepoClean.value
+    stageAndCacheKey.value
```
jkinkead: Why don't you just use the task in `VersionInjectorPlugin` directly here? I don't see why you have a new task for this . . .

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -40,6 +44,18 @@ object VersionInjectorPlugin extends AutoPlugin {
       "gitDescribe",
       "The description of the current git commit"
     )
+    val gitMRC = TaskKey[String](
```
jkinkead: This name should match the other hash-generating task (`gitSha1`). Maybe `gitLocalSha1`?

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -40,6 +44,18 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+      "gitMRC",
+      "Most recent commit in src"
+    )
+    val cacheKey = TaskKey[String](
+      "cacheKey",
+      "cacheKey of project"
+    )
+    val mTask = TaskKey[Seq[String]](
+      "fucksbt",
```
jkinkead: I agree with the sentiment, but this should probably be removed. :P

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -72,6 +91,14 @@ object VersionInjectorPlugin extends AutoPlugin {
     (gitCommand("describe").!!).trim
   }
 
+  def gitMostRecentCommit(path: File) = {
```
jkinkead: Return type?

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -72,6 +91,14 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+    gitCommand("log", "-1", "--format=%H", path).!!.trim
+  }
+
+  val gitMRCTask = gitMRC := {
+    gitMostRecentCommit(sourceDirectory.value)
```
jkinkead: This should use the project base (`baseDirectory.value`). It's possible that some other files (like `build.sbt`) will affect the output jar - and the number of extra misses due to this will be fairly small.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +144,45 @@ object VersionInjectorPlugin extends AutoPlugin {
         gitConfFile
     }
 
+  val injectCacheKeyTask =
+    injectCacheKey <<= (resourceManaged in Compile, organization, name, version, streams, cacheKey) map {
```
jkinkead: Is there a reason you can't use `:=` and call `.value` on the dependencies?

I ask because I honestly can't figure out how `<<=` differs from `:=`, and I'd prefer to use `:=` if possible.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +144,45 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+      // scalastyle: ignore
+      (resourceManaged, org, name, version, s, cacheKeyResult) =>
+        val cacheKeyConfFile = resourceManaged / org / cleanArtifactName(name) / "cacheKey.conf"
+        val cacheKeyContents = "cacheKey: " + "\"" + cacheKeyResult + "\""
```
jkinkead: Alternately:
```scala
// Triple-quotes with interpolation:
s"""cacheKey: "$cacheKeyResult""""
```

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
ryanai3: I was just following the style of injectGitTask, and injectArtifactTask. I'm sure I could use :=, but I'm not sure how it would work

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +144,45 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+        IO.write(cacheKeyConfFile, cacheKeyContents)
+        cacheKeyConfFile
+    }
+
+  lazy val gitMRCs = Def.taskDyn {
```
jkinkead: This should be commented. In particular, you should describe what it's doing (finding  the most recent commits for all local dependencies), and why it needs a dynamic task.

jkinkead: A type on this value would also be good.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -14,7 +14,7 @@ object VersionInjectorPlugin extends AutoPlugin {
   object autoImport {
     val injectVersion = TaskKey[Seq[File]](
       "injectVersion",
-      "Generates both the artifact.conf and git.conf version files"
+      "Generates the artifact.conf, git.conf, and cacheKey.conf version files"
```
jkinkead: Revert - this doesn't generate the cache key anymore.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +144,45 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+      case _ => Seq.empty
+    }
+    val stageDir = baseDirectory.value + "/target/universal/stage"
+    val allFiles: Seq[JFile] = Seq("bin", "conf", "lib", "public")
+      .map { dir => stageDir + "/" + dir }
+      .flatMap(getFileList)
+
+    val (filesToGetGitSha, filesToHash) = allFiles partition { f: JFile =>
+      val fileName = f.getName
+      fileName.startsWith("org.allenai.ari-api") ||
```
jkinkead: You should either have a TODO to make this a setting, or go ahead and add a setting for ignored local jars.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +144,45 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+        fileName.startsWith("org.allenai.solvers-") ||
+        fileName.startsWith("org.allenai.ari-solvers-")
+    }
+    val hashes = filesToHash.map(Hash.apply).map(Hash.toHex).mkString
```
jkinkead: This still relies on filesystem and SBT subproject ordering, both of which could fluctuate. You should put all hash values into a set, sort them, then take the final hash value.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
jkinkead: I *think* `<<=` just says "I want the values of all of these keys". You should just be able to call `.value` on all of the keys in the body of `:=`.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
ryanai3: on injectCacheKeyTask, or on gitMRC's?

jkinkead: `gitMRC` should have a type declaration - it's not clear from reading what it is.

ryanai3:
@jkinkead made changes, passes manual tests

build.sbt
```diff
@@ -36,7 +36,7 @@ addSbtPlugin(
 addSbtPlugin("com.github.jkinkead" % "sbt-scalariform" % "0.1.6")
 
 // Dependency graph visualiztion in SBT console
-addSbtPlugin("com.gilt" % "sbt-dependency-graph-sugar" % "0.7.4")
+addSbtPlugin("com.gilt" % "sbt-dependency-graph-sugar" % "0.7.5-1")
```
jkinkead: If this isn't required by your plugin, you may want to remove (or put in a separate PR).

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -122,6 +124,8 @@ object DeployPlugin extends AutoPlugin {
     IO.delete((UniversalPlugin.autoImport.stagingDirectory in Universal).value)
   }
 
+  val stageAndCacheKeyTask = stageAndCacheKey := VersionInjectorPlugin.injectCacheKeyTask
```
jkinkead: Remove? This is no longer needed.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -40,6 +44,14 @@ object VersionInjectorPlugin extends AutoPlugin {
       "gitDescribe",
       "The description of the current git commit"
     )
+    val gitLocalSha1 = TaskKey[String](
+      "gitLocalSha1",
+      "Most recent commit in src"
+    )
+    val cacheKey = TaskKey[String](
+      "cacheKey",
+      "cacheKey of project"
```
jkinkead: You should expand this a bit (and capitalize the first word for consistency with the other keys).

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -72,6 +87,14 @@ object VersionInjectorPlugin extends AutoPlugin {
     (gitCommand("describe").!!).trim
   }
 
+  def gitMostRecentCommit(path: File): String = {
+    gitCommand("log", "-1", "--format=%H", path).!!.trim
```
jkinkead: I'd move this into the task def below - it's not used elsewhere, so a separate helper isn't really needed. Maybe add a comment to it, though?

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +140,51 @@ object VersionInjectorPlugin extends AutoPlugin {
         gitConfFile
     }
 
+  val injectCacheKeyTask = injectCacheKey := {
+    val values = ((resourceManaged in Compile).value.absolutePath, organization.value, name.value, streams.value, cacheKey.value)
```
jkinkead: I'd split these out into separate lines.
```scala
val resourceManagedPath = (resourceManaged in Compile).value.absolutePath
// etc.
```

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +140,51 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+    val (resourceManagedPath, org, nameValue, s, cacheKeyResult) = values
+    val cacheKeyConfFile = new java.io.File(s"$resourceManagedPath/$org/${cleanArtifactName(nameValue)}/cacheKey.conf")
+    s.log.info(s"Generating cacheKey.conf managed resource... (cacheKey: $cacheKeyResult)")
```
jkinkead: More common is to do:
```scala
val logger = streams.value.log
logger.info("stuff")
```

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +140,51 @@ object VersionInjectorPlugin extends AutoPlugin {
         gitConfFile
     }
 
+  val injectCacheKeyTask = injectCacheKey := {
+    val values = ((resourceManaged in Compile).value.absolutePath, organization.value, name.value, streams.value, cacheKey.value)
+    val (resourceManagedPath, org, nameValue, s, cacheKeyResult) = values
+    val cacheKeyConfFile = new java.io.File(s"$resourceManagedPath/$org/${cleanArtifactName(nameValue)}/cacheKey.conf")
```
jkinkead: I'd make `cleanName` a local.
```scala
val cleanName = cleanArtifactName(name.value)
```

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +140,51 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+    s.log.info(s"Generating cacheKey.conf managed resource... (cacheKey: $cacheKeyResult)")
+
+    val cacheKeyContents = s"""cacheKey: "$cacheKeyResult""""
+    IO.write(cacheKeyConfFile, cacheKeyContents)
+    cacheKeyConfFile
+  }
+
+  //Gives us the git most recent commits for all the local projects that this project depends on
```
jkinkead: Space after `//`.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +140,51 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+  lazy val gitMRCs = Def.taskDyn { // We have to use a dynamic task because generating tasks based on project dependencies
```
jkinkead: I'd put this comment up above, with the other comment on the task.

jkinkead: Also, you should rename to something clearer, and add a type.
```scala
lazy val dependentGitCommits: Seq[Task[String]] // or whatever this type is
```

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +140,51 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+    // necessarily alters the task graph
+    val MRCs = buildDependencies.value.classpathRefs(thisProjectRef.value).sorted // get the local dependencies
```
jkinkead: It would be better to sort the SHA1 values, since you know that they have a stable sort order. I'm not sure what the type here is, but it looks like an sbt internal type that may change ordering on you.

Also, you probably just want to sort at the very end - the ordering in this function isn't really important . . .

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +140,51 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+    val filter = ScopeFilter(inProjects(MRCs: _*)) // this is weird, we create a scopefilter on the dependencies
+    gitLocalSha1.all(filter) // odd piece of syntax - returns a list of tasks (we're inside a taskDyn block) that is
+    //applying gitLocalSHa1 to all scopes in the scopefilter 'filter'
+  }
+
+  val cacheKeyTask = cacheKey := {
+    import java.io.{ File => JFile }
```
jkinkead: I think you can just use sbt's builtin - it has implicit conversions . . .

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +140,51 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+
+    def getFileList(dir: String): Seq[JFile] = new JFile(dir) match {
```
jkinkead: Did you try using the path matchers I linked you to? You should be able to one-line most of this helper . . .

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +140,51 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+      //TODO: make this a setting
+      val fileName = f.getName
+      fileName.startsWith("org.allenai.ari-api") ||
+        fileName.startsWith("org.allenai.solvers-") ||
+        fileName.startsWith("org.allenai.ari-solvers-")
+    }
+    val hashes = filesToHash.sorted
+      .map(Hash.apply)
+      .map(Hash.toHex)
+    Hash.toHex(Hash((hashes ++ gitMRCs.value :+ gitLocalSha1.value).sorted.mkString))
```
jkinkead: This is the only place you need to have a `sorted` call. You should add a comment, too, so that the sorting doesn't get deleted by an over-zealous refactorer later on . . .

jkinkead:
More comments!

ryanai3:
More fixes!

build.sbt
jkinkead: Ping!

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
jkinkead: Ping!

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +136,51 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+
+    val cacheKeyContents = s"""cacheKey: "$cacheKeyResult""""
+    IO.write(cacheKeyConfFile, cacheKeyContents)
+    cacheKeyConfFile
+  }
+
+  //Gives us the git most recent commits for all the local projects that this project depends on
+  // We have to use a dynamic task because generating tasks based on project dependencies
+  lazy val dependentGitCommits: Def.Initialize[Task[Seq[String]] = Def.taskDyn {
+    // necessarily alters the task graph
```
jkinkead: Move up to the first half of this comment.

Also, you should terminate your sentences with a period.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -117,6 +136,51 @@ object VersionInjectorPlugin extends AutoPlugin {
...
+    val MRCs = buildDependencies.value.classpathRefs(thisProjectRef.value)// get the local dependencies
```
jkinkead: Prefer comments above the line they're documenting to at the end-of-line.

jkinkead:
Awesome, this is starting to look really great! Let me know when you've fixed up your comments - I think this will be good to merge after that. :)

You should consider working on the TODO in your next PR (making the filtered jars a setting).

ryanai3:
Fixed comments, Merging!

build.sbt
```diff
@@ -36,7 +36,7 @@ addSbtPlugin(
 addSbtPlugin("com.github.jkinkead" % "sbt-scalariform" % "0.1.6")
 
 // Dependency graph visualiztion in SBT console
-addSbtPlugin("com.gilt" % "sbt-dependency-graph-sugar" % "0.7.4")
+addSbtPlugin("com.gilt" % "sbt-dependency-graph-sugar" % "0.7.4444")
```
jkinkead: You didn't fix this - is this a valid version?

PR #175, opened 2015-07-09
run-class.sh now adds cacheKey conf file to jvm config var

ryanai3:


ryanai3:
@jkinkead is my bash correct?

jkinkead:
If you had an integration test, you wouldn't need to ask me this . . . :P

src/main/resources/run-class.sh
```diff
@@ -98,9 +98,17 @@ if [ -e conf/env.conf ]; then
   CONF_FILE="-Dconfig.file=conf/env.conf"
 fi
 
+#Add a cache-key config, user tests for existence
+ADD_CACHE_KEY=""
```
jkinkead: You have three variables, all with similar names: `ADD_CACHE_KEY`, `ADD_CACHEKEY`, and `ADDCACHEKEY`. For best results, use just one. :)

src/main/resources/run-class.sh
```diff
@@ -98,9 +98,17 @@ if [ -e conf/env.conf ]; then
...
+if [ -e conf/cacheKey.conf ]; then
+  CACHEKEY_FILE_CONTENTS = $(<cacheKey.conf)
+  CACHEKEY=${CACHEKEY_FILE_CONTENTS##*: }
```
jkinkead: If you're just catting to an environment variable, you should fix the file contents to be the SHA value directly, so that you don't need janky bash regex magic.

src/main/resources/run-class.sh
```diff
@@ -98,9 +98,17 @@ if [ -e conf/env.conf ]; then
...
+  ADD_CACHEKEY="-Dapplication.cacheKey=$CACHEKEY"
+fi
+
 CLASSPATH=`find lib -name '*.jar' | tr "\\n" :`
 JAVA_CMD=(java $JVM_ARGS -classpath $CLASSPATH $CONF_FILE
-  ${LOGBACK_CONF[@]})
+  ${LOGBACK_CONF[@]} $ADDCACHEKEY)
```
jkinkead: You should put your variable in double-quotes (although you'd also need to fix your regex to drop the quotes in your conf file).

jkinkead:
Yeah, I'd definitely go with creating a `.sha1` file with only the SHA1 contents, instead of parsing out the config value.

ryanai3:
@jkinkead pls review

ryanai3:
@jkinkead ping

ryanai3:
@jkinkead pingerino

src/main/resources/run-class.sh
```diff
@@ -98,9 +98,16 @@ if [ -e conf/env.conf ]; then
   CONF_FILE="-Dconfig.file=conf/env.conf"
 fi
 
+#Add a cache-key config, user tests for existence
+ADD_CACHE_KEY=""
+if [ -e conf/cacheKey.Sha1 ]; then
+  CACHEKEY = $(<cacheKey.conf)
```
jkinkead: Fix.

src/main/resources/run-class.sh
```diff
@@ -98,9 +98,16 @@ if [ -e conf/env.conf ]; then
...
+  ADD_CACHE_KEY="-Dapplication.cacheKey=$"CACHEKEY""
```
jkinkead: Fix - extra quotes around CACHEKEY.

src/main/resources/run-class.sh
ryanai3: done

src/main/resources/run-class.sh
ryanai3: done

src/main/resources/run-class.sh
```diff
@@ -98,9 +98,16 @@ if [ -e conf/env.conf ]; then
   CONF_FILE="-Dconfig.file=conf/env.conf"
 fi
 
+#Add a cache-key config, user tests for existence
+ADD_CACHE_KEY=""
+if [ -e conf/cacheKey.Sha1 ]; then
+  CACHEKEY = $(<cacheKey.Sha1)
```
jkinkead: I don't think this assignment will work with spaces. I get the feeling you're not testing this . . .

jkinkead:
One more comment.

src/main/resources/run-class.sh
ryanai3: fixed... automated testing is on the way.

jkinkead: I'm mostly concerned that this doesn't look to be *manually* tested, either. ;)

jkinkead:
LGTM!

PR #177, opened 2015-07-24
stageAndCacheKey defined incorrectly - taskKey accidentally capitaliz…

ryanai3:
…ed, pure function not instantiated in plugin


jkinkead:
Sweet! Merging.

PR #178, opened 2015-08-20
correct directory we find cacheKey in

ryanai3:
@jkinkead - Quick fix

### DIFF
--- a/src/main/resources/run-class.sh
+++ b/src/main/resources/run-class.sh
@@ -100,7 +100,7 @@ fi
 
 #Add a cache-key config, user tests for existence
 ADD_CACHE_KEY=""
-if [ -e conf/cacheKey.Sha1 ]; then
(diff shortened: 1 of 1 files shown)


src/main/resources/run-class.sh
```diff
@@ -100,7 +100,7 @@ fi
 
 #Add a cache-key config, user tests for existence
 ADD_CACHE_KEY=""
-if [ -e conf/cacheKey.Sha1 ]; then
+if [ -e ../conf/cacheKey.Sha1 ]; then
```
jkinkead: This isn't correct. See line 56 - this changes to the directory above `bin`, which contains `conf`. Note that all of the other configuration paths use `conf` as the base . . .

ryanai3:
you're right, something was messy on my end

PR #179, opened 2015-08-20
look for cacheKey.Sha1 in correct place...

ryanai3:
@jkinkead thx for the help


jkinkead:
LGTM! I'll work on publishing.

PR #181, opened 2015-09-01
Fix DeployPlugin to auto-ignore the dependent jars.

jkinkead:
Diffs in these are handled by the dependent git commits.

This fixes a TODO that was making a bunch of spurious cache misses.


ryanai3:
LGTM!

PR #184, opened 2015-09-29
Fix the cache key to filter out transitive local project dependencies…

jkinkead:
… instead of just direct local project dependencies.

This fixes the case where we have a complicated local project tree, where a local project is depended on transitively but not directly - something like `frontend` depends on `frontend_api` which depends on `backend_common`.

@ryanai3 - FYI. Oh, the joys of sbt. :)


row11:
LGTM, merging

ryanai3:
@jkinkead - SBT makes *some* things easy :D

PR #185, opened 2015-10-01
Scaladoc generation plugin, with support for publishing to github pages.

jkinkead:
This is much simpler than it looks; the vast majority of this code is merely to support linking out from our Scaladoc to that of other projects we depend on.

**Important Note**: This introduces a bug if our plugins are loaded from a build that's not rooted at the base of a git repository. Aristo doesn't do this at all, except in the tester projects in `sbt-plugins`. I don't believe S2 (or any other team) does this, but if they do, there's a workaround in the plugin.


docs/scaladoc-gen.md
```diff
@@ -0,0 +1,39 @@
...
+```scala
+// Enable the plugin on your project, which should aggregate all of your
+// subprojects.
+lazy val myRootProject = project.in(file(".")).aggregate(one, two)
+  .settings(
+    // This setting is required in order for the plugin to load.
+    scaladocGenGitRemoteRepo := "git@github.com:allenai/myprojectname.git",
+    // This is not strictly required, but is needed if you want your library to
+    // be linkable by other projects in their scaladoc.
+    apiURL := Some(url("https://allenai.github.io/myprojectname/latest/api/"))
```
markschaake: so what does this do exactly? Does it generate a link in the scaladocs home page or something?

src/main/scala/org/allenai/plugins/ScaladocGenPlugin.scala
```diff
@@ -0,0 +1,216 @@
...
+  /** Setting that overrides the doc generation task. This finds all HTML files generated by the doc
+    * task, then subs in the correctly-formatted Javadoc link in place of the badly-formatted
+    * Scaladoc link.
+    */
+  lazy val fixJavadocSetting = doc in ScalaUnidoc := {
+    // Build up a regular expression that matches any of the external javadoc links we have.
+    val externalUrls =
+      scaladocGenJavadocUrl.value +: scaladocGenExtraJavadocMap.value.values.toVector
+    val anyUrl = externalUrls map { url => s"""\\Q$url\\E""" } mkString { "|" }
+    val captureUrlAndFragment = s""""($anyUrl)#([^"]*)"""".r
```
markschaake: could you add a comment showing an example fully-baked regex? Would be easier to parse what's going on here.

src/main/scala/org/allenai/plugins/ScaladocGenPlugin.scala
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  // The following are helpers to make it easier to add manual mappings for Scala libraries.
+
+  lazy val scaladocGenExtraScaladocMap: SettingKey[Map[String, URL]] = settingKey[Map[String, URL]](
+    "Mappings for any Scala libraries to link in, of library name to Scaladoc URL"
+  )
+  /** The default setting includes libraries from CoreDependencies. */
+  lazy val extraScaladocMapSetting = scaladocGenExtraScaladocMap := {
+    val akkaUrl = url("http://doc.akka.io/api/akka/2.4.0")
```
markschaake: does this mean that scaladocs are hard-coded against a particular version and not resolved based on actual project dependency versions? E.g. if one subproject overrides the akka version, will its generated scaladocs link to the wrong akka version docs?

docs/scaladoc-gen.md
jkinkead: According to the [sbt docs](http://www.scala-sbt.org/0.13/docs/Howto-Scaladoc.html), this embeds something in the POM that clients can read when generating docs, so that you don't need to manually configure API mappings.

jkinkead: Digging more, it adds a `properties` element with the name `info.apiURL` and the API URL as the value. I believe only sbt looks for this.

markschaake:
The amount of code and additional sbt plugin dependencies is a bit scary, and hardcoded URL versions (java / akka / etc.) seem brittle. It is also surprising that this much ceremony is required to get scaladocs generated for multi-project build. Oh well, SBT is SBT.

The fact this doesn't have an impact on runtime and that it is an opt-in plugin mitigates my concerns... so LGTM and merge when ready!

src/main/scala/org/allenai/plugins/ScaladocGenPlugin.scala
jkinkead: Yes, without overriding this map setting for the appropriate key(s). Note that if we sent akka a patch that added the `info.apiURL` property to their pom.xml, we could remove these. That's (obviously) true of any of these. I might do that real quick . . .

src/main/scala/org/allenai/plugins/ScaladocGenPlugin.scala
jkinkead: Done!

jkinkead:
In a few ways, the ugliness is the result of the scaladoc compiler - but yeah, sbt is sbt. I don't love the hard-coded versions, but sbt does provide a way to get rid of them, if we can get the open-source projects to do it. :)

PR #186, opened 2015-10-02
Add in the SBT site mappings.

jkinkead:
Hah, I kept remembering this was missing, but kept forgetting to add it . . . :)

I also forgot to test the non-hacked version of the plugin on a real github pages.

### DIFF
--- a/src/main/scala/org/allenai/plugins/ScaladocGenPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/ScaladocGenPlugin.scala
@@ -217,8 +217,8 @@ object ScaladocGenPlugin extends AutoPlugin {
         // This will pull in links from pom.xml files that sbt has generated, when the project sets
         // the apiURL key.
         autoAPIMappings := true,
-        // This adds the output of Unidoc to the SbtSite plugin's site directory (what gets synced
-        // to github pages).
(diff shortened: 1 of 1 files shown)


src/main/scala/org/allenai/plugins/ScaladocGenPlugin.scala
```diff
@@ -216,7 +216,10 @@ object ScaladocGenPlugin extends AutoPlugin {
         scalaApiMappingsSetting,
         // This will pull in links from pom.xml files that sbt has generated, when the project sets
         // the apiURL key.
-        autoAPIMappings := true
+        autoAPIMappings := true,
+        // This adds the output of Unidoc to the SbtSite plugin's site directory (what gets synced
+        // to github pages).
```
markschaake: Is this producing a site index or something? I don't know these terms so can't grok the comment here.

jkinkead: `sbt-site` just copies files into `target/site`, based on mappings. It's like the universal plugin's `stage` command.

`sbt-ghpages` then takes this directory and copies it to the `gh-pages` branch, then pushes it to github.

This is saying "take the mappings unidoc provides for the site plugin, and put the output in "latest/api"".

jkinkead:
I tried to clarify the comment a bit.

markschaake:
LGTM

PR #187, opened 2015-10-13
Don't assume that "moduleId" is set.

jkinkead:
sbt docs don't provide info on when or why this would or wouldn't be set. It's been set always on other projects I've tested, but aristo causes problems - my best guess is that the jar file in the tableilp solver doesn't have this key set.

Regardless, this fixes this plugin for aristo.


sbhaktha:
LGTM!

PR #190, opened 2015-11-18
Upgrade common dependencies to 1.0.19

markschaake:
@jkinkead here's the other half - I'll release on merge.


jkinkead:
Easy! :)

Issue #194, opened 2015-12-20
Support execution of multiple scripts with npm:build


rkrzewski:
I'm using jspm and gulp for building my application's frontend. When the application is build from scratch (notably in CI build), `jspm install` needs to be performed before `gulp build` is started. I was fiddling around with different ways I could wire this together and the cleanest I was able to come up so far was declaring two separate `script` entries in `package.json` for the two tasks. 

When trying to integrate my gulp based build with SBT, I came across your plugin which I like a lot - it's orders of magnitude simpler that sbt-web ;) However, `npm:build` has a limitation that the invoked script is hardcoded to `build`. I was able to work around this by defining a third `script` entry in `package.json` that invokes a script that spawns two instances of `npm` in sequence, to execute the two other scripts. 

I'd like to propse a new setting for the plugin `nodeBuildScripts` of type `Seq[String]` with default value `Seq("build")`. `execBuild` method would then iterate over this sequence and run each of the requested scripts. The change would be fully backward compatible.

markschaake:
@rkrzewski glad you are interested in using our plugins! You should be able to override the `build in Npm` task to do what you want it to do via something like:

```scala
// in build.sbt

NodeKeys.build in Npm := {
  // do stuff I need to happen before the default build task:
  ...
  // now run the previously set (default) build task:
  (NodeKeys.build in Npm).value
}
```

I recommend trying the above approach before issuing a PR. We use this plugin on several internal projects at AI2, and a change coming from an external PR would require someone on our team to do some manual testing before approval. This means it could take some time before someone can take a good look at the PR - probably not until Jan 2016. But please do issue a PR if you think it would be a better solution than what I've suggested above!

rkrzewski:
January 2016 sounds great :) I have a JS based workaround right now, and secondly I am just prototyping a solution that would allow me to migrate a project from sbt-web to a native nodejs tooling.

markschaake:
Resolved via #195

PR #195, opened 2015-12-22
Allow specifying multiple scripts to be executed with npm:build task

rkrzewski:
Addresses issue #194

### DIFF
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -158,6 +158,7 @@ object DeployPlugin extends AutoPlugin {
       Def.task {
         NodeJsPlugin.execBuild(
           (NodeKeys.nodeProjectDir in Npm).value,
+          (NodeKeys.buildScripts in Npm).value,
           NodeJsPlugin.getEnvironment("dev", (NodeKeys.nodeProjectTarget in Npm).value),
           (NodeKeys.npmLogLevel in Npm).value
         )
@@ -166,6 +167,7 @@ object DeployPlugin extends AutoPlugin {
       Def.task {
         NodeJsPlugin.execBuild(
           (NodeKeys.nodeProjectDir in Npm).value,
+          (NodeKeys.buildScripts in Npm).value,
           NodeJsPlugin.getEnvironment("prod", (NodeKeys.nodeProjectTarget in Npm).value),
           (NodeKeys.npmLogLevel in Npm).value
         )
--- a/src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/NodeJsPlugin.scala
@@ -45,7 +45,9 @@ object NodeJsPlugin extends AutoPlugin {
     }
 
     object NodeKeys {
-      val build = taskKey[Seq[File]]("Execution `npm run build` in the Node application directory")
+      val build = taskKey[Seq[File]](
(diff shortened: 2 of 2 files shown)


rkrzewski:
The change itself is quite trivial, but I guess it should come together with a test anyway.

I can provide a test project that uses jspm and gulp (and thus needs `npm install` followed by two distinct `npm run` executions to build from scratch). Would you like me to emulate the existing `test-node-js` project, or any minimal project will do?

I also have a question regarding the build process of `sbt-plugin` project itself. I see `"org.scala-sbt" % "scripted-plugin"` in project dependencies, but I cannot see any configuration for it in `build.sbt`. I'd expect to find something because your test projects are in different location that `src/sbt-test` assumes. Also, when I invoke `scripted` task in SBT the test projects are _not_ processed. However, when I invoke `test` they are, but I really can't see how that happens. I'm not asking out of idle curiosity, but because that I cannot see any assertions being made in the tests. Does the testing process really checks only that each of test projects builds succeeds? I think I might be missing something important here...

markschaake:
> Does the testing process really checks only that each of test projects builds succeeds? I think I might be missing something important here...

I actually don't know how our tests work either. The original intent was to use sbt-scripted, but we had an intern (who is no longer here) move forward with a different test strategy. I think ultimately we want to use sbt-scripted (since that is the standard way to test SBT plugins), but for now we are not. I'd say for your PR, making another test project is good enough, and someone on our end can manually verify before merging.

markschaake:
Your changes look great to me. I'll plan on testing it and merging my first day back from the holiday break.

Thanks for contributing!

jkinkead:
It's my understanding that the tests (by @ryanai3, who could chime in) are just assertions that the plugins load correctly - but there might be something more to it. IIRC there was a reason why we didn't use sbt-scripted, but I don't remember what that reason was. :)

ryanai3:
@jkinkead no problem :). There's plugin load assertions, but I think I went crazy and made sure that cachekeys changed on dependencies but didn't change when they were supposed to, etc.

So we decided to use scalatest (which was in use for all the other scala testing at the time) instead of scripted because we'd get the same sort of 'unit testing for scripts' with less code. There's a test [here](https://github.com/allenai/sbt-plugins/blob/master/src/test/scala/CacheKeyTestSpec.scala) you can look at.

There's not all that much documentation on scripted I could find at the time if you don't want to trudge through code. Although it's simple, that was one of the factors - scalatest is a lot more well-documented. There's [this](http://eed3si9n.com/testing-sbt-plugins) and [this](http://stackoverflow.com/questions/19102208/how-to-test-sbt-plugins). 

Scripted makes it really easy to mess around with the filesystem and make sure that your plugin responds correctly, but you still have to write scala code for more complex checking of behavior, unless your plugin does everything through the filesystem and the names of files.

You'll note in the first link that in order to have custom assertions or checking in tests, you need to write it as an sbt task anyway - and then call it from scripted and check equality there. 

Given that, we thought it would be just as good to write the tests in a sort of BDD format with scalatest instead of writing the tests as sbt tasks, and calling them from scripted. It also allows us to do some more complex testing that might involve running the sbt command in different directories (because sbt gets annoying with the 'i'm recursive... but not really').

markschaake:
I'm reviewing this PR now

markschaake:
Local testing done. LGTM.

markschaake:
New release 1.1.7 published: https://bintray.com/allenai/sbt-plugins/allenai-sbt-plugins/1.1.7/view

PR #199, opened 2016-01-07
Change `sed` syntax to fix CI

markschaake:
Linux doesn't like a space after the `-i` flag in `sed`. This removes that space so our build can pass on Semaphore.

@jkinkead quick review? This will get us a green badge on our repo landing page.


jkinkead:
Huh. LGTM!

PR #200, opened 2016-01-07
Remove dead test code

markschaake:
@jkinkead one more. This is removing dead code that resulted from hastily splitting apart the `simple` and `cache-key` tests.


jkinkead:
LGTM!

PR #205, opened 2016-01-20
Upgrade common dependency to latest release

markschaake:
Also:

- Make it simpler to override common release version for an entire project

@jkinkead last one today!


vha14:
LGTM

PR #206, opened 2016-01-28
Make formatter not format on compile.

jkinkead:
This fixes the old annoyance cropping up again. :)

I also added a unit test for it, and confirmed that this fixes the issue.


markschaake:
LGTM!

PR #207, opened 2016-02-04
Generate autoformatting git pre-commit hook via task

markschaake:
- Adds an SBT task `generateAutoformatGitHook` that copies scalariform.jar and pre-commit resources from our plugins to the client project's .git/hooks directory.
- Makes sure our scalariform default overrides are consistent between SBT and the commit hook by auto-generating scalariform CLI options
- Fails fast if the user has conflicting files in place
- Added a simple check that the files are generated as expected. I also created a local test project and verified the hooks work. I don't have time right now to add the scripted tests for this - but would be good to add later (that the hooks will cause actual change in format for source files).

@jkinkead and @milescrawford I think this will make onboarding a bit easier. This does not address the "global template" feature, but maybe that is becoming less important as our projects at AI2 as we grow.

@jkinkead could you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,6 +36,8 @@ addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "0.8.0" excludeAll(
   ExclusionRule(organization = "com.danieltrinh")
 ))
 
+// If you change the scalariform version, you may also need to generate a new
+// scalariform.jar in src/main/resources/autoformat/
 addSbtPlugin("org.scalariform" % "sbt-scalariform" % "1.6.0")
 
 // Dependency graph visualiztion in SBT console
--- a/src/main/resources/autoformat/pre-commit
+++ b/src/main/resources/autoformat/pre-commit
@@ -1,0 +1,91 @@
+#!/bin/sh
+
+# Git pre-commit hook to check staged scala files for formatting issues.
+#
+# This requires that scalariform is installed alongside the pre-commit script, as scalariform.jar.
+# This can be built with "sbt cli/oneJar" - see the README for more details.
+#
+# This script has one gitconfig parameter, hooks.autoformat, which controls how badly-formatted
+# files are handled.
+#
+# Default mode:
+# When autoformat is off, this will test all staged files, and will exit with an error code if any
+# are misformatted, printing the files with errors.
+#
+# Autoformat mode:
+# When autoformat is on, this first checks for unstaged changes to staged files, and if there are
+# any, it will exit with an error. Files with unstaged changes will be printed.
+#
+# If all staged files have no unstaged changes, it will run scalariform against them, leaving the
+# formatting changes unstaged. Changed files will be printed.
+
+# Find all staged scala files, and exit early if there aren't any.
+SCALA_FILES=(`git diff --name-only --cached --diff-filter=AM | grep --color=never '.scala$'`)
+if [ ! "$SCALA_FILES" ]; then
+  exit 0
+fi
+
+# Validate that the scalariform jar exists.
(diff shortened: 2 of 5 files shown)


src/main/resources/autoformat/pre-commit
```diff
@@ -0,0 +1,91 @@
...
+# Validate that the scalariform jar exists.
+JAR=`dirname $0`/scalariform.jar
+if [ ! -e $JAR ]; then
+  echo "ERROR: $JAR not found!"
+  exit 2
+fi
+
+# The formatting options we change from the defaults. Values are set by the CoreSettingsPlugin
+# when running generateAutoformatGitHook task.
+OPTIONS=__scalariform_opts__
```
markschaake: The only difference between this file and what @jkinkead wrote is that the options are injected by the plugin to ensure consistent default overrides between SBT and the git pre-commit hook.

milescrawford:
looks rad, and makes sense to associate with the project, not the global ~/.git_template

build.sbt
```diff
@@ -36,6 +36,8 @@ addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "0.8.0" excludeAll(
   ExclusionRule(organization = "com.danieltrinh")
 ))
 
+// If you change the scalariform version, you may also need to generate a new
+// scalariform.jar in src/main/resources/autoformat/
```
jkinkead: You ought to be able to pull this out of the ivy cache - but good luck figuring out how! :)

Maybe add a TODO?

src/main/scala/org/allenai/plugins/CoreSettingsPlugin.scala
```diff
@@ -141,6 +150,71 @@ object CoreSettingsPlugin extends AutoPlugin {
...
+      val preCommitFile = expectedGitHooksDir / "pre-commit"
+      val scalariformFile = expectedGitHooksDir / "scalariform.jar"
+      def requireFilesDontExist(files: File*) = {
+        files find (_.exists) foreach { file =>
+          sys.error(s"You already have .git/hooks/${file.getName}. Remove or rename the file and run again.")
+        }
+      }
+      requireFilesDontExist(preCommitFile, scalariformFile)
+
+      def withResource[A](resource: String)(f: InputStream => A) = {
```
jkinkead: What about using sbt's `IO.readLinesURL` and `IO.writeLines` instead of writing these helpers?

jkinkead:
A couple of suggestions.

src/main/scala/org/allenai/plugins/CoreSettingsPlugin.scala
markschaake: I swear I looked for those! Thanks, I'll make the change. BTW, did you know that the `IO` is gone from sbt master? Have you seen that SBT has split into two repos: one for 0.13 and one for 1.0?

0.13: https://github.com/sbt/sbt-zero-thirteen
1.0: https://github.com/sbt/sbt/

jkinkead: I knew that they were making big changes for `1.0.*`, but I didn't realize they'd moved the old branches to a new repo!

In particular, they're trying to modularize some of the really tangled code that's in there. My guess is that `IO` got moved to a utils project (I think it was in a utils subproject in `0.13.X`).

I'm hoping they'll make the compiler more modular - it's super-hard to get it to work as-is as a standalone!

build.sbt
markschaake: Hmmm, I think the scalariform CLI is a different jar than what shows up in the ivy cache, since the scalariform jar is 1.9 M and the CLI (fat assembled jar) is ~ 8 M.

PR #208, opened 2016-02-09
Allow custom args to be passed through to main class in run-class.sh.

danxmoran:
Required for specifying custom solver ports in aristo

PR #209, opened 2016-02-16
Pass optional args to restart script on deploy, refactor in prep for deploying to many hosts.

danxmoran:
A few things here:

- Separate validation of environment variables from validation of configuration
- Create dedicated `case class` wrappers for deploy config
- Lay the groundwork for deploying to multiple hosts in parallel
- Add support for a new config key, `deploy.startup_args`.  If this config entry is set to an array of strings, the contents will be passed as arguments to the call to `deploy.startup_script`, making use of #208 
- Resolve a bunch of IntelliJ style warnings

### DIFF
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -11,8 +11,11 @@ import sbt.Keys._
 import java.io.File
 
 import scala.collection.JavaConverters._
+import scala.concurrent.{ Await, Future }
+import scala.concurrent.duration.Duration
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.sys.process.{ Process, ProcessLogger }
-import scala.util.Try
+import scala.util.{ Failure, Success, Try }
 
 /** Plugin to deploy a project to an ec2 instance. Handles copying binaries and config files,
   * generating a basic startup script, copying over an environment-specific config file if it
@@ -47,7 +50,7 @@ object DeployPlugin extends AutoPlugin {
   override def requires: Plugins = plugins.JvmPlugin && JavaAppPackaging && NodeJsPlugin
 
   /** Static usage string. */
-  val Usage = "Usage: deploy [overrides] [deploy target]";
+  val Usage = "Usage: deploy [overrides] [deploy target]"
 
   object autoImport {
 
@@ -86,7 +89,7 @@ object DeployPlugin extends AutoPlugin {
 
   /** sbt.Logger wrapper that prepends [deploy] to log messages */
   case class DeployLogger(sbtLogger: Logger) {
-    private def logMsg(msg: String) = s"[deploy] ${msg}"
+    private def logMsg(msg: String) = s"[deploy] $msg"
     def info(msg: String): Unit = sbtLogger.info(logMsg(msg))
     def error(msg: String): Unit = sbtLogger.error(logMsg(msg))
   }
@@ -187,26 +190,30 @@ object DeployPlugin extends AutoPlugin {
     if (reducedArgs.length != 1) {
       throw new IllegalArgumentException(Usage)
     }
+
+    // Ensure necessary environment variables are set.
+    validateEnv()
+
     val workingDirectory = (baseDirectory in thisProject).value
     val configFile = new File(workingDirectory.getPath + "/conf/deploy.conf")
-    if (!configFile.isFile()) {
-      throw new IllegalArgumentException(s"${configFile.getPath()}: Must be a config file")
+    if (!configFile.isFile) {
+      throw new IllegalArgumentException(s"${configFile.getPath}: Must be a config file")
     }
 
-    val deployTarget = reducedArgs(0)
+    val deployTarget = reducedArgs.head
 
     val targetConfig = loadTargetConfig(commandlineOverrides, configFile, deployTarget)
 
-    val configMap = validateAsMap(deployTarget, targetConfig)
+    val deployConfig = parseConfig(deployTarget, targetConfig)
 
     // TODO(jkinkead): Allow for a no-op / dry-run flag that only prints the
     // commands.
 
     // Check out the provided version, if it's set.
-    for (version <- configMap.get("project.version")) {
-      log.info(s"Checking out ${version} . . .")
+    deployConfig.projectVersion foreach { version =>
+      log.info(s"Checking out $version . . .")
       if (Process(Seq("git", "checkout", "-q", version)).! != 0) {
-        throw new IllegalArgumentException(s"Could not checkout ${version}.")
+        throw new IllegalArgumentException(s"Could not checkout $version.")
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -221,56 +228,79 @@ object DeployPlugin extends AutoPlugin {
...
+        val restartCommand = Seq.concat(
+          sshCommand,
+          Seq(deployHost, restartScript, "restart"),
+          hostConfig.startupArgs map { "--" +: _ } getOrElse Seq()
+        )
+        log.info("Running " + restartCommand.mkString(" ") + " . . .")
+        if (Process(restartCommand).! != 0) {
+          throw new IllegalArgumentException(s"Error running restart command on host '$deployHost'.")
+        }
+      } map Success.apply recover { case e => Failure(e) } // Map to Try so we can fold over errors.
```
jkinkead: It would be clearer to just wrap the body of the `Future.apply` call in `Try.apply`.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -221,56 +228,79 @@ object DeployPlugin extends AutoPlugin {
...
     }
 
-    // Command to pass to rsync's "rsh" flag, and to use as the base of our ssh
-    // operations.
-    val sshCommand = {
-      val pemEnvVar = "AWS_PEM_FILE"
-      val sshKeyfile = sys.env(pemEnvVar)
-      val sshUser = configMap("deploy.user.ssh_username")
-      Seq("ssh", "-i", sshKeyfile, "-l", sshUser)
+    val deploys = deployConfig.hostConfigs map { hostConfig =>
```
jkinkead: It's helpful to declare the type when it's a little nonobvious:
```scala
val deploys: Iterable[Future[Try[Unit]]]
```

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -221,56 +228,79 @@ object DeployPlugin extends AutoPlugin {
...
-    // Shell-friendly version of rsync command, with rsh value quoted.
-    val quotedRsync = rsyncCommand.patch(
-      2, Seq("--rsh=" + sshCommand.mkString("\"", " ", "\"")), 1
-    ).mkString(" ")
-    log.info("Running " + quotedRsync + " . . .")
-    if (Process(rsyncCommand).! != 0) {
-      throw new IllegalArgumentException("Error running rsync.")
+
+    // Accumulate all error messages thrown when trying to deploy.
+    val accumulatedErrors = Future.fold[Try[Unit], Option[String]](deploys)(None) {
```
jkinkead: This is a bit confusing.

Also, I'd accumulate the full list of exceptions, then build a string later:
```scala
val errors = Future.fold[Try[Unit], Seq[Throwable]](deploys)(Seq.empty) {
  case (acc, Success(_)) => acc
  case (acc, Failure(e)) => acc :+ e
}
```

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -420,51 +449,95 @@ object DeployPlugin extends AutoPlugin {
...
     */
-  def validateAsMap(targetName: String, targetConfig: Config): Map[String, String] = {
-    // Validate that the target has all the required environment variables.
+  def validateEnv(): Unit = {
     val requiredEnv = Seq("AWS_PEM_FILE")
-    for (variable <- requiredEnv) {
-      val value = sys.env.get(variable).getOrElse {
-        throw new IllegalArgumentException(s"Environment variable ${variable} undefined.")
-      }
+    requiredEnv foreach { variable =>
```
jkinkead: I'm actually pretty sure this isn't needed for deploys anymore.

jkinkead: That is, we have passwordless ssh configured for all of our boxes now, so this mechanism is obsolete.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -420,51 +449,95 @@ object DeployPlugin extends AutoPlugin {
...
     }
+  }
+
+  /** Container for config needed to deploy a project to a host.
+    * @param host The fully-qualified name of the host the project should be deployed to
+    * @param directory The directory on the host in which the project's code should be placed
+    * @param sshUser The username to use when accessing the host
+    * @param startupScript The path, relative to the deploy directory, to the script that should
+    *                      be used to restart the project after deploying the new code
+    * @param startupArgs Optional arguments to pass to the startup script
```
jkinkead: Just make an `Seq`? You can always check that for emptiness.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -420,51 +449,95 @@ object DeployPlugin extends AutoPlugin {
...
-    val requiredKeys = Seq("deploy.host", "deploy.directory",
-      "deploy.startup_script", "deploy.user.ssh_username")
-    val requiredKeyPairs: Seq[(String, String)] = for {
-      key <- requiredKeys
-    } yield key -> {
-      try {
-        targetConfig.getString(key)
+    DeployConfig(
+      hostConfigs = Seq(parseHostConfig(targetConfig.getConfig("deploy"))),
+      projectVersion = Try(targetConfig.getString("project.version")).toOption
```
jkinkead: I don't think that's what you want. Typesafe config shouldn't throw an exception if a key is missing - it should return `null`.

You ought to be able to do `Option(targetConfig.getString("project.version"))`.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -420,51 +449,95 @@ object DeployPlugin extends AutoPlugin {
...
+    )
+  }
+
+  /** Transform the given [[com.typesafe.config.Config]] object into a corresponding
+    * [[org.allenai.plugins.DeployPlugin.HostConfig]] object.
+    * @param hostConfig host-level config to convert
+    */
+  def parseHostConfig(hostConfig: Config): HostConfig = {
+    lazy val renderedConfig = hostConfig.root().render()
```
jkinkead: I would not bother making this lazy.

jkinkead:
Comments!

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: `getString` does throw a `ConfigException.Missing` if the requested value is missing or null: http://typesafehub.github.io/config/latest/api/com/typesafe/config/Config.html#getString-java.lang.String-

danxmoran: That said, I probably want to modify this to be more granular and let any `ConfigException.WrongType` errors bubble through

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: Should I remove it?

danxmoran:
Comments addressed!

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -420,51 +456,87 @@ object DeployPlugin extends AutoPlugin {
...
-    val requiredEnv = Seq("AWS_PEM_FILE")
-    for (variable <- requiredEnv) {
-      val value = sys.env.get(variable).getOrElse {
-        throw new IllegalArgumentException(s"Environment variable ${variable} undefined.")
+  def parseConfig(targetName: String, targetConfig: Config): DeployConfig = {
+
+    DeployConfig(
+      hostConfigs = Seq(parseHostConfig(targetConfig.getConfig("deploy"))),
+      projectVersion = try {
+        Some(targetConfig.getString("project.version"))
```
danxmoran: I mentioned on a now-outdated diff, the `getString` method does throw an exception instead of returning `null` if a requested path doesn't have a value: http://typesafehub.github.io/config/latest/api/com/typesafe/config/Config.html#getString-java.lang.String-

src/main/scala/org/allenai/plugins/DeployPlugin.scala
jkinkead: I must be spoilt by the Scala API.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
jkinkead: I'd leave this in for now, unless you want to verify the daily deploy without it. I think the evaluation framework may still rely on it, since it's using our old AMI.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -221,56 +225,89 @@ object DeployPlugin extends AutoPlugin {
...
-    // Command to pass to rsync's "rsh" flag, and to use as the base of our ssh
-    // operations.
-    val sshCommand = {
-      val pemEnvVar = "AWS_PEM_FILE"
-      val sshKeyfile = sys.env(pemEnvVar)
-      val sshUser = configMap("deploy.user.ssh_username")
-      Seq("ssh", "-i", sshKeyfile, "-l", sshUser)
+    val deploys: Seq[Future[Try[Unit]]] = deployConfig.hostConfigs map { hostConfig =>
+      Future {
+        // Run commands within a Try so we can fold over any errors.
```
jkinkead: You may want to be more clear: `so we can collect all errors`?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -221,56 +225,89 @@ object DeployPlugin extends AutoPlugin {
...
-    // Now, ssh to the remote host and run the restart script.
-    val restartScript = deployDirectory + "/" + configMap("deploy.startup_script")
-    val restartCommand = sshCommand ++ Seq(deployHost, restartScript, "restart")
-    log.info("Running " + restartCommand.mkString(" ") + " . . .")
-    if (Process(restartCommand).! != 0) {
-      throw new IllegalArgumentException("Error running restart command.")
+    // Throw an error if any deploy commands failed.
+    val deployStatus = accumulatedErrors map { errorSeq =>
+      if (errorSeq.nonEmpty) {
+        throw new IllegalArgumentException(
```
jkinkead: This is not a good type for this exception - the error could be anything.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -221,56 +225,89 @@ object DeployPlugin extends AutoPlugin {
...
+          "Hit the following errors during deployment:" +
+            (errorSeq map { _.getMessage } mkString ("\n", "\n", "\n"))
```
jkinkead: I personally find the three-arg `mkString` a little awkward.

I'd probably do:
```scala
val errors = errorSeq.mkString("\n")
throw new Exception(s"Hit the following errors during deployment:\n$errors\n")
```
I'd also probably format more like `(err, err, err)` instead of putting newlines in - they are a bit hard-to-read, and don't make it obvious these came from a list.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -420,51 +456,87 @@ object DeployPlugin extends AutoPlugin {
     overrides.withFallback(rcConfig.withFallback(targetConfig)).resolve
   }
 
-  /** Translates the given deploy Config object into a flat Map[String, String]
-    * containing all the relevant deployment keys.
-    *
-    * Exits the program if any of the keys required for deployment are missing.
-    *
+  /** Container for config needed to deploy a project to a host.
+    * @param host The fully-qualified name of the host the project should be deployed to
```
jkinkead: Lower-case first word after argument name in (Scala|Java)doc.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -420,51 +456,87 @@ object DeployPlugin extends AutoPlugin {
...
+  }
 
-      require(!value.isEmpty, "Environment variable ${value} is empty.")
-    }
+  /** Transform the given [[com.typesafe.config.Config]] object into a corresponding
+    * [[org.allenai.plugins.DeployPlugin.HostConfig]] object.
+    * @param hostConfig host-level config to convert
+    */
+  def parseHostConfig(hostConfig: Config): HostConfig = {
+    val renderedConfig = hostConfig.root().render()
```
jkinkead: You can also inline this below where it's needed. It's only in two places - and it's only two function calls.

jkinkead: Oh, three places. Still.

jkinkead:
A few small suggestions.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: `IllegalArgumentException` is abused pretty heavily throughout this file.  I'll look into more useful exception types.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -420,51 +456,87 @@ object DeployPlugin extends AutoPlugin {
...
+        hostConfig.getString(key)
       } catch {
-        case configError: ConfigException =>
-          configError match {
-            case _: ConfigException.Missing =>
-              throw new IllegalArgumentException(s"Error: ${targetName} missing key ${key}.")
-            case _: ConfigException.WrongType =>
-              throw new IllegalArgumentException(s"Error: ${targetName}.${key} must be a string.")
-          }
+        case _: ConfigException.Missing =>
```
danxmoran: Should we just let these exceptions bubble up rather than catching them and converting them to `IllegalArgumentException`s?  This is about as informative of an exception type as we'll be able to get without reinventing the wheel here.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
jkinkead: It's mostly reasonable if it's indicating an error in the input files. It's not reasonable if it's an error from a deploy.

danxmoran: What about the exceptions thrown for missing environment variables / for git status problems?  Would `IllegalStateException` be more appropriate?

jkinkead: Yeah, those sound fine.

There's actually a way to register errors in sbt - I think it's the `error` method, but I'm not sure. You should use that instead.

This code started life as a scala script outside of sbt - so it's a bit confusing. :)

danxmoran: The sbt docs show examples of using `error` but I can't find it defined in their code, so I'm assuming it's just the `error` method from Scala's `Predef` object, which has been deprecated in favor of `sys.error`.  The method is just:
```Scala
def error(message: String): Nothing = throw new RuntimeException(message)
```
So I'll use it here but keep more-specific exception types where appropriate.

danxmoran:
Ready for another look!

jkinkead:
LGTM! I assume you've tested?

danxmoran:
Yup! Working fine deploying locally both with startup_args and without.

danxmoran:
Semaphore doesn't seem to want to test my latest commit...

jkinkead:
It did that once to me recently. I couldn't get it to run the correct commit.

danxmoran:
It's working fine deploying to localhost, and "sbt test" passes.  I'll let you pull the trigger on the merge if you're comfortable ignoring Semaphore's strange behavior.

jkinkead:
This could be due to the DDoS attack Hiram mentioned.

I'll merge!

PR #210, opened 2016-02-18
Add full support for deploying to many replicas in parallel.

danxmoran:
Implementation of the spec described in part 1 of [my design plan](https://docs.google.com/document/d/1CBKdHb7KDYHmquPEmH3uhPlWVtH79NyxiTTc1F0m34s/edit#).

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -126,3 +126,6 @@ the args to the call to `servicename.sh`, separated from the `start|restart` com
 
     $ ./servicename.sh start -- arg1 arg2 --flag
     $ ./servicename.sh restart -- other1 --otherflag
+
+This is not recommended for normal deploys, however, since it makes it impossible to manually
+restart the service with the same start state without looking into deploy configuration.
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -82,6 +82,117 @@ A simple configuration using `global_deploy.conf` could look like:
       }
     }
 
+Running the deploy task with this config would result in a single instance of `servicename`
+being deployed to the host `ari-staging.allenai.org`, with code placed in the remote directory
+`/local/deploy/servicename`.
+
+A more complex configuration would involve deploying multiple instances of a host in parallel. An
+example configuration achieving this would be:
+
+    staging = {
+      include "global_deploy.conf"
+
+      project = {
+        name = "servicename"
+      }
+      deploy = {
+        replicas = [
+          {
+            host = "ari-staging.allenai.org"
+          }
+          {
+            host = "ari-staging.allenai.org"
+          }
+        ]
+      }
+    }
+
+Deploying with this config would set up two instances of `servicename` on `ari-staging.allenai.org`,
+with code placed in the remote directories `/local/deploy/servicename-1` and
+`/local/deploy/servicename-2`.
+
+The naming scheme of remote directories the plugin deploys to is determined by the number of
+replicas being deployed to a given host. If only one replica is being deployed to a host, then
+only one directory will be created on the host, and it will be named according to the value of
+the `deploy.directory` key in config (which in turn can be derived from the `project.name` key).
+On the other hand, if multiple replicas are being deployed to a host then multiple directories
+will be created on the host; the names of these directories will take the form `deploy.directory-i`,
+where `i` is the index of the replica contained in the directory.
+
+The above example is contrived in that it deploys two identical services on the same host. More
+likely, we'd want to deploy two instances of a service with slightly different configuration (to
+prevent conflict over port bindings, for example). The deploy task supports this usage by
+allowing per-replica config overrides to be specified in `deploy.conf`, for example:
+
+    staging = {
+      include "global_deploy.conf"
+
+      project = {
+        name = "servicename"
+      }
+      deploy = {
+        replicas = [
+          {
+            host = "ari-staging.allenai.org",
+            config_overrides = { port = 2000 }
+          }
+          {
+            host = "ari-staging.allenai.org"
+            config_overrides = { port = 3000 }
+          }
+        ]
+      }
+    }
+
+Note that overrides are merged directly with the top-level object defined by the service's
+environment configuration (for example, `dev.conf`), so keys must match for overrides to take hold.
+It's also possible to specify a `config_overrides` key in the top-level `deploy` object. Any
+overrides listed in that location will be applied to all replicas, or to the single top-level
+host if only one service instance is being deployed.
+
+The deploy task will handle stopping and cleaning up any replicas that become stale as deploy
+configuration changes over time. For example, if a project with config:
+
+    staging = {
+      include "global_deploy.conf"
+
+      project = {
+        name = "servicename"
+      }
+      deploy = {
+        replicas = [
+          {
+            host = "ari-staging.allenai.org",
+            config_overrides = { port = 2000 }
+          }
+          {
+            host = "ari-staging.allenai.org"
+            config_overrides = { port = 3000 }
+          }
+        ]
+      }
+    }
+
+Was modified to have config:
+
+    staging = {
+      include "global_deploy.conf"
+
+      project = {
+        name = "servicename"
+      }
+      deploy = {
+        host = "ari-staging.allenai.org"
+      }
+    }
+
+On the next deploy, the deploy task will stop the old replicas running in
+`/local/deploy/servicename-1` and `/local/deploy/servicename-2` and delete those directories
+before deploying to `/local/deploy/servicename`. The same would be true in reverse. The task
+would not, however, delete the two directories if a third replica was added instead of the second
+being removed, since the directories would still be valid for use according to the directory
+naming scheme outlined above.
+
 The `ari-core` project has a more complicated example using more features of the [HOCON language](https://github.com/typesafehub/config/blob/master/HOCON.md).
 
 ### caching
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -120,3 +120,9 @@ The deploy plugin will now stage this file to the `bin` directory with run-class
     $ ./servicename.sh restart # Same as 'stop' followed by 'start'.
 
 This will also use `servicename` for your logback appname.
+
+If you need to pass arguments to the service's main class on startup, you can do so by appending
+the args to the call to `servicename.sh`, separated from the `start|restart` command by `--`:
+
+    $ ./servicename.sh start -- arg1 arg2 --flag
+    $ ./servicename.sh restart -- other1 --otherflag
--- a/src/main/resources/global_deploy.conf
+++ b/src/main/resources/global_deploy.conf
@@ -15,12 +15,9 @@ project = {
   version = null
 }
 deploy = {
-  // Hostname to push to. Required.
+  // Hostname to push to. Required if 'replicas' not set.
   host = null
-  // Directory on the remote host to push to. Required.
-  directory = "/local/deploy/"${?project.name}
-  // Start / stop script to run after the push is complete. Required.
-  startup_script = "bin/"${?project.name}".sh"
+  // User data for the host above. Required if 'host' is set.
   user = {
     // Full path to the ssh keypair to use when connecting to the remote host.
     // Required.
@@ -28,4 +25,20 @@ deploy = {
     // Username to connect to the remote host as. Required.
     ssh_username = "ec2-user"
   }
+  // Directory on the remote host to push to. Required.
+  directory = "/local/deploy/"${?project.name}
+  // Start / stop script to run after the push is complete. Required.
+  startup_script = "bin/"${?project.name}".sh"
+  // Optional arguments to pass to the startup script. Forbidden if 'host' is not set.
+  startup_args = []
+  // An array of objects describing replicas to push to. Required if 'host' not set.
+  // Objects take the form
+  // {
+  //   // Required.
+  //   host = null
+  //   user = { ssh_keyfile = null, ssh_username = null }
+  //   // Optional.
+  //   startup_args = []
+  // }
+  replicas = null
 }
--- a/src/main/resources/global_deploy.conf
+++ b/src/main/resources/global_deploy.conf
@@ -15,9 +15,7 @@ project = {
   version = null
 }
 deploy = {
-  // Hostname to push to. Required if 'replicas' not set, forbidden if 'replicas' is set.
-  host = null
-  // User data for the host above. Required if 'host' is set.
+  // User data to use during deployment. Required.
   user = {
     // Full path to the ssh keypair to use when connecting to the remote host.
     // Required.
@@ -29,17 +27,20 @@ deploy = {
   directory = "/local/deploy/"${?project.name}
   // Start / stop script to run after the push is complete. Required.
   startup_script = "bin/"${?project.name}".sh"
+
+  // Hostname to push to. Required if 'replicas' not set, forbidden if 'replicas' is set.
+  host = null
   // An array of objects describing replicas to push to. Required if 'host' not set, forbidden
   // if 'host' is set.
   // Objects take the form
   // {
   //   // Required.
   //   host = null
-  //   user = { ssh_keyfile = null, ssh_username = null }
   //   // Optional. Only applies to this replica.
   //   config_overrides = {}
   // }
   replicas = null
+
   // Optional configuration overrides to apply to the 'application.conf' files of all replicas
   // on deploy.
   config_overrides = {}
--- a/src/main/resources/global_deploy.conf
+++ b/src/main/resources/global_deploy.conf
@@ -15,7 +15,7 @@ project = {
   version = null
 }
 deploy = {
-  // Hostname to push to. Required if 'replicas' not set.
+  // Hostname to push to. Required if 'replicas' not set, forbidden if 'replicas' is set.
   host = null
   // User data for the host above. Required if 'host' is set.
   user = {
@@ -29,16 +29,18 @@ deploy = {
   directory = "/local/deploy/"${?project.name}
   // Start / stop script to run after the push is complete. Required.
   startup_script = "bin/"${?project.name}".sh"
-  // Optional arguments to pass to the startup script. Forbidden if 'host' is not set.
-  startup_args = []
-  // An array of objects describing replicas to push to. Required if 'host' not set.
+  // An array of objects describing replicas to push to. Required if 'host' not set, forbidden
+  // if 'host' is set.
   // Objects take the form
   // {
   //   // Required.
   //   host = null
   //   user = { ssh_keyfile = null, ssh_username = null }
-  //   // Optional.
-  //   startup_args = []
+  //   // Optional. Only applies to this replica.
+  //   config_overrides = {}
   // }
   replicas = null
+  // Optional configuration overrides to apply to the 'application.conf' files of all replicas
+  // on deploy.
+  config_overrides = {}
 }
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -1,7 +1,7 @@
 package org.allenai.plugins
 
 import org.allenai.plugins.NodeJsPlugin.autoImport.{ NodeKeys, Npm }
-import com.typesafe.config.{ Config, ConfigException, ConfigFactory, ConfigValueFactory }
+import com.typesafe.config._
 import com.typesafe.sbt.SbtNativePackager._
 import com.typesafe.sbt.packager.archetypes.JavaAppPackaging
 import com.typesafe.sbt.packager.universal.UniversalPlugin
@@ -62,7 +62,9 @@ object DeployPlugin extends AutoPlugin {
         "Defaults to false"
     )
 
-    val filterNotCacheKeyGenFileNames = settingKey[Seq[String]]("starts of jars in stage you don't want hashed for cachekey")
+    val filterNotCacheKeyGenFileNames = settingKey[Seq[String]](
+      "Starts of jars in stage you don't want hashed for cachekey"
+    )
 
     val cleanStage = taskKey[Unit](
       "Cleans the staging directory. This is not done by default by the universal packager."
@@ -76,7 +78,7 @@ object DeployPlugin extends AutoPlugin {
       */
     val deployDirs = settingKey[Seq[String]](
       "subdirectories from the stage task to copy during deploy. " +
-        "defaults to bin/, conf/, lib/, and public/"
+        "defaults to bin/, lib/, and public/"
     )
 
     val gitRepoClean = taskKey[Unit]("Succeeds if the git repository is clean")
@@ -225,10 +227,9 @@ object DeployPlugin extends AutoPlugin {
     val universalStagingDir = stageAndCacheKey.value
 
     val envConfFile = new File(universalStagingDir, s"conf/$deployEnv.conf")
-    if (envConfFile.exists) {
-      log.info(s"Copying config for $deployEnv . . .")
-      val destConfFile = new File(universalStagingDir, "conf/env.conf")
-      IO.copyFile(envConfFile, destConfFile)
+    val baseEnvConf = if (envConfFile.exists) {
+      log.info(s"Getting base config for $deployEnv . . .")
+      ConfigFactory.parseFile(envConfFile)
     } else {
       log.info("")
       log.info(s"WARNING: Could not find config file '$deployEnv.conf'!")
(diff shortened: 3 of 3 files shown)


danxmoran:
Not sure what's up with the build failure...
```
Could not create file /tmp/sbt_2f8795d6/cache-key/webservice/target/streams/$global/ivyConfiguration/$global/streams/outjava.io.IOException: No such file or directory
```

docs/deploy.md
```diff
@@ -120,3 +120,9 @@ The deploy plugin will now stage this file to the `bin` directory with run-class
     $ ./servicename.sh restart # Same as 'stop' followed by 'start'.
 
 This will also use `servicename` for your logback appname.
+
+If you need to pass arguments to the service's main class on startup, you can do so by appending
+the args to the call to `servicename.sh`, separated from the `start|restart` command by `--`:
+
+    $ ./servicename.sh start -- arg1 arg2 --flag
```
jkinkead: Rethinking this a little: I don't think it's a good idea to require flags to be passed to these. If we require things like `--port` to be specified on the commandline, it means that we can't start or stop services in production without looking into our deployment configuration.

We should look into either using environment variables or system properties to pass in the port (which typesafe config supports), or look into auto-generating these start / stop scripts from a template.

:/

danxmoran:
I've updated this to use `config_overrides` as we talked about yesterday.

docs/deploy.md
danxmoran: This is addressed, but I left this in the README because it's still supported.

danxmoran:
Made one more addition to check for, stop, & remove any replicas that won't be updated as part of a deploy run before trying to update / restart the replicas listed in config.

Ready for another review pass!

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -76,7 +78,7 @@ object DeployPlugin extends AutoPlugin {
       */
     val deployDirs = settingKey[Seq[String]](
       "subdirectories from the stage task to copy during deploy. " +
-        "defaults to bin/, conf/, lib/, and public/"
```
danxmoran: Rather than copying the whole of `conf/`, this now loads the appropriate environment configuration (i.e. `dev.conf`) in a way that fully resolves imports and substitutions, applies overrides to that `Config` object, then writes it alone to the replica's directory as `conf/env.conf`.

docs/deploy.md
jkinkead: Maybe add a note that it's not recommended for normal deploys, since it's not easily reproduceable?

src/main/resources/global_deploy.conf
```diff
@@ -15,12 +15,9 @@ project = {
...
 }
 deploy = {
-  // Hostname to push to. Required.
+  // Hostname to push to. Required if 'replicas' not set, forbidden if 'replicas' is set.
   host = null
-  // Directory on the remote host to push to. Required.
-  directory = "/local/deploy/"${?project.name}
-  // Start / stop script to run after the push is complete. Required.
-  startup_script = "bin/"${?project.name}".sh"
+  // User data for the host above. Required if 'host' is set.
```
jkinkead: Is the intent that this will be used for all hosts? It should be - think should normally never change.

src/main/resources/global_deploy.conf
```diff
@@ -28,4 +25,22 @@ deploy = {
...
+  directory = "/local/deploy/"${?project.name}
+  // Start / stop script to run after the push is complete. Required.
+  startup_script = "bin/"${?project.name}".sh"
+  // An array of objects describing replicas to push to. Required if 'host' not set, forbidden
+  // if 'host' is set.
+  // Objects take the form
+  // {
+  //   // Required.
+  //   host = null
+  //   user = { ssh_keyfile = null, ssh_username = null }
```
jkinkead: See above. This should probably not be here.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
jkinkead: I would not change this - it's quite possible we're relying on other items in `conf/` being copied over.

jkinkead: More clearly: I would continue to sync all `conf` contents. Updating how `env.conf` is populated sounds awesome.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -236,64 +237,170 @@ object DeployPlugin extends AutoPlugin {
...
+      // as part of the deploy.
+      val (deployParent, deployTarget) = baseDeployDir.splitAt(baseDeployDir.lastIndexOf('/'))
+      // Strip leading / from deployTarget so we can match names on it.
+      val namePattern = deployTarget.tail
+      // Build a regex matching the replicas we will eventually update / restart as part of this
+      // process, and therefore DON'T want to stop and remove.
+      val restartedPattern = if (numReplicas == 1) {
+        namePattern
+      } else {
+        s"$namePattern-[1-$numReplicas]"
```
jkinkead: So, I don't think we want *all* deploys to include a replica number - only those that deploy multiple times to the same host.

This might be too complicated to try to write logic for. Specifying all of the directories seems better than having to worry about which items are needing to be restarted or not.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -236,64 +237,170 @@ object DeployPlugin extends AutoPlugin {
...
+      }
+      // Use find -exec to locate, stop, and remove stale replicas that won't be updated as part
```
jkinkead: This scares me a little.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -236,64 +237,170 @@ object DeployPlugin extends AutoPlugin {
...
+        case (replicaConfig, i) => {
+          Future {
+            // Determine the specific remote directory for this replica.
+            val deployDirectory = if (numReplicas > 1) {
+              s"$baseDeployDir-${i + 1}"
+            } else {
+              baseDeployDir
+            }
+
+            // Command to pass to rsync's "rsync-path" flag to ensure creation of the remote dir.
```
jkinkead: This seems over-complicated.

jkinkead: That is, I'm not sure why you're not just running `mkdir`.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -236,64 +237,170 @@ object DeployPlugin extends AutoPlugin {
...
+
+      log.info("Running " + quotedStopCommand.mkString(" ") + " . . .")
+      if (Process(stopCommand).! != 0) {
+        sys.error(s"Error while trying to tear down stale replicas on '$deployHost'.")
+      }
 
-          log.info("Running " + restartCommand.mkString(" ") + " . . .")
-          if (Process(restartCommand).! != 0) {
-            sys.error(s"Error running restart command on host '$deployHost'.")
+      replicas.zipWithIndex map {
```
jkinkead: This loop is doing a lot of things, and it's not immediately clear what those are. Maybe comment and / or break into smaller helper functions?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -236,64 +237,170 @@ object DeployPlugin extends AutoPlugin {
...
+              ),
+              rsyncDirs,
+              Seq(
+                "--exclude=/*",
+                "--delete",
+                universalStagingDir.getPath + "/",
+                deployHost + ":" + deployDirectory
+              )
+            )
+            // Shell-friendly version of rsync command, with rsh value quoted.
```
jkinkead: This is confusing.

It's being quoted so you can print it out & allow copy-pasting, yes? You may wish to comment on that.

It would also be less confusing if this were created closer to use.

jkinkead:
I'm having a bit of trouble following all of the commands in the middle of your function.

I'd suggest putting the command construction behind a helper method, maybe with clear docs? Many of the fields are constants; you could also have a template `Seq`.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: Some of the complexity of this block comes from the fact that we don't want to include a replica number for deploys pointing at only one host.  My main worry was the case when you jump from pointing at one remote to pointing at > 1 remote.

For example, we're running one `inference-solver` in dev right now.  On the host machine there's a directory `/local/deploy/inference-solver`, and we don't want to change that unless we add a second instance.  When we do add a second instance, the directory names will become `/local/deploy/inference-solver-1.sh` and `/local/deploy/inference-solver-2.sh`.  Presumably the config for the first solver (which already existed) won't change, which means the solver in `inference-solver-1` will be trying to bind to the same port as the solver running in `inference-solver`.  The existing restart commands won't handle stopping the solver in `inference-solver` because they only run in the newly-deployed directories, and the deploy will fail when `inference-solver-1` fails to bind to its port.

danxmoran: Also potential clarification needed: `numReplicas` here is relative to a single host (it's running within a `flatMap` over hosts).

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: Me too, but it works like a charm when deploying locally :D

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: The `--rsync-path` is just a <del>trick</del> hack to make the `mkdir` and `rsync` run in a single remote call.

I do seem to have a tendency to over-optimize.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: Can do!

docs/deploy.md
danxmoran: Note added below!

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -86,6 +88,48 @@ object DeployPlugin extends AutoPlugin {
 
   import autoImport._
 
+  override def projectSettings: Seq[Def.Setting[_]] = Seq(
```
danxmoran: Moved this up so it's not mixed with the deploy helpers.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -177,34 +221,59 @@ object DeployPlugin extends AutoPlugin {
     }
   }
 
+  // We stage and generate the cache key in the same task so that stage only runs once.
```
danxmoran: Also moved this so it's not lost in the deploy logic.

danxmoran:
Comments are addressed!  Each system command now has its own documented helper method.

danxmoran:
I've verified that the current version of changes works when deploying to EC2 instances in the dev VPC, both for deploying many replicas to a single host and for deploying one replica to each of many hosts.

jkinkead:
I'm still concerned about the complexity . . . but I don't want to hold you up, and my brain's too fuzzy today to make concrete suggestions.

Can you be sure to update the readme to clearly document how each replica case is handled?

danxmoran:
Can do!  This also isn't blocking my controller additions anymore now that it's in a state where I can use it to shard solvers out to my own test EC2 instances, so there isn't a huge rush to merge.

danxmoran:
@jkinkead do you have time to look at this again today?  I'm not sure I'll hit my goal of having the dev controller running against a sharded setup by the meeting on Monday because of thread madness, but I'd at least like to run this new logic through a few nightly deploys before then

docs/deploy.md
```diff
@@ -82,6 +82,117 @@ A simple configuration using `global_deploy.conf` could look like:
...
+Deploying with this config would set up two instances of `servicename` on `ari-staging.allenai.org`,
+with code placed in the remote directories `/local/deploy/servicename-1` and
+`/local/deploy/servicename-2`.
+
+The naming scheme of remote directories the plugin deploys to is determined by the number of
+replicas being deployed to a given host. If only one replica is being deployed to a host, then
+only one directory will be created on the host, and it will be named according to the value of
+the `deploy.directory` key in config (which in turn can be derived from the `project.name` key).
+On the other hand, if multiple replicas are being deployed to a host then multiple directories
+will be created on the host; the names of these directories will take the form `deploy.directory-i`,
```
jkinkead: `$i` in the pattern might be more readable.

docs/deploy.md
```diff
@@ -82,6 +82,117 @@ A simple configuration using `global_deploy.conf` could look like:
...
     }
 
+Running the deploy task with this config would result in a single instance of `servicename`
+being deployed to the host `ari-staging.allenai.org`, with code placed in the remote directory
+`/local/deploy/servicename`.
+
+A more complex configuration would involve deploying multiple instances of a host in parallel. An
+example configuration achieving this would be:
+
+    staging = {
```
jkinkead: FWIW you can also enclose a block in triple-` to get code formatting. It's usually slightly easier than messing with indent.

jkinkead:
Thanks for being patient, looking at this now. The README updates are awesome!

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -236,64 +300,79 @@ object DeployPlugin extends AutoPlugin {
...
-        // Now, ssh to the remote host and run the restart script.
-        val restartScript = deployDirectory + "/" + hostConfig.startupScript
-        val restartArgs = if (hostConfig.startupArgs.nonEmpty) {
-          "--" +: hostConfig.startupArgs
-        } else {
-          Seq()
-        }
+    // Command to pass to rsync's "rsh" flag, and to use as the base of our ssh operations.
+    val sshCommand = {
+      val keyFile = sys.env.getOrElse("AWS_PEM_FILE", {
```
jkinkead: Now that my brain's working again, I think this can be made optional - and we just omit the `-i` flag from `ssh` if it's missing. Most users should have passwordless ssh set up for the prod deploy user. Maybe add a TODO note about this?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -236,64 +300,79 @@ object DeployPlugin extends AutoPlugin {
...
-        // Run commands within a Try so we can collect errors using Future.fold.
-        // TODO(danm): Use ProcessIO to capture and report the actual stderr of failures here.
-        Try {
-          log.info("Running " + quotedRsync + " . . .")
-          if (Process(rsyncCommand).! != 0) {
-            sys.error(s"Error running rsync to host '$deployHost'.")
+    // Loop over all unique remote hosts listed in config, deploying to them in parallel.
+    val hosts = deployConfig.replicaOverridesByHost.keys
+    val deploys: Iterable[Future[Try[Unit]]] = hosts flatMap { hostName =>
+      val baseDeployDir = deployConfig.baseDirectory
```
jkinkead: For these global settings, you should initialize outside of the loop.

jkinkead: Although this may be the only global one. :)

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -236,64 +300,79 @@ object DeployPlugin extends AutoPlugin {
...
+
+      val replicas = deployConfig.replicaOverridesByHost(hostName)
```
jkinkead: Move closer to first use.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -459,92 +486,306 @@ object DeployPlugin extends AutoPlugin {
...
+    // Build a pattern matching the replica directories that we want to maintain for subsequent
+    // updates. This pattern will be passed to `find` as a negated name match.
+    val persistedPattern = if (numReplicas == 1) {
+      // If only one replica is going to be deployed, we want to keep the deploy directory
+      // without a numeric suffix.
+      projectName
+    } else {
+      // If multiple replicas are going to be deployed, we want to keep one deploy directory for
+      // each of them.
+      s"$projectName-[1-$numReplicas]"
```
jkinkead: This doesn't work if `numReplicas` is > 9.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -459,92 +486,306 @@ object DeployPlugin extends AutoPlugin {
...
+    // replicas.
+    val findCommand = Seq(
+      "find",
+      // Prevent `find` from reporting subdirectories of the deployed replicas.
+      deployParent, "-maxdepth", "1",
+      // Set `find` to only report directory names.
+      "-type", "d",
+      // Set `find` to only report directories named after this project (to avoid cleaning other
+      // deployed projects).
+      "-name", "\"" + projectName + "*\"",
```
jkinkead: If you want to use interpolation with double-quotes, you can use a triple-quoted interpolated string:
```scala
s""""$projectName*""""
```
It's weird that the parser accepts this, but it does :)

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -459,92 +486,306 @@ object DeployPlugin extends AutoPlugin {
...
+      // After stopping the replica service, delete it from the deploy directory.
+      "-exec", "rm -r {} \\;"
     )
+
+    // Build the SSH command that will be sent to the remote host to execute the above `find`.
+    val teardownCommand = (sshCommand :+ host) ++ findCommand
+    // Also build a shell-friendly version of the command to log out, for copy-paste.
+    val quotedTeardownCommand = quoteSshCmd(sshCommand, host, findCommand)
+
+    // Log and run the command on the remote host, and throw an error if it fails.
```
jkinkead: There should be a TODO to not consider this an error based on a flag. Maybe.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -236,64 +300,79 @@ object DeployPlugin extends AutoPlugin {
...
-            sys.error(s"Error running rsync to host '$deployHost'.")
+    // Loop over all unique remote hosts listed in config, deploying to them in parallel.
+    val hosts = deployConfig.replicaOverridesByHost.keys
+    val deploys: Iterable[Future[Try[Unit]]] = hosts flatMap { hostName =>
+      val baseDeployDir = deployConfig.baseDirectory
+
+      val replicas = deployConfig.replicaOverridesByHost(hostName)
+      val numReplicas = replicas.length
+
+      // First, ensure parent directory structure of deploy exists.
```
jkinkead: Typically you won't have permission to create this (`/local` usually owned by `root`). Verifying it exists is nice, though.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -459,92 +486,306 @@ object DeployPlugin extends AutoPlugin {
...
+      ),
+      rsyncIncludes,
+      Seq(
+        "--exclude=/*",
+        "--delete",
+        srcDir,
+        s"$host:$targetDir"
+      )
+    )
+    // Also build a shell-friendly version of the command to log out, for copy-paste.
```
jkinkead: Use helper method?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -459,92 +486,306 @@ object DeployPlugin extends AutoPlugin {
...
+
+    // Command to print the full env config, to be piped to SSH.
+    val echoConfCommand = Seq("echo", renderedConf)
+    // Shell-friendly version of above, for copy-paste.
+    val quotedEcho = Seq("echo", s"'$renderedConf'").mkString(" ")
+
+    // SSH command to receive `echo` results and redirect them to disk.
+    val writeConfCommand = sshCommand :+ host :+ s"cat > $confPath/env.conf"
+    // Shell-friendly version of above, for copy-paste.
+    val quotedWrite = quoteSshCmd(sshCommand, host, Seq(s"cat > $confPath/env.conf"))
```
jkinkead: quote-log-exec-fail is done about four times in this file - maybe factor that into a helper instead of `quoteSshCmd`?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -459,92 +486,306 @@ object DeployPlugin extends AutoPlugin {
...
+    host: String,
+    envConf: Config,
+    confPath: String,
+    sshCommand: Seq[String],
+    log: DeployLogger
+  ): Unit = {
+    // Get the config contents to write.
+    val renderedConf = envConf.root().render(ConfigRenderOptions.concise)
+
+    // Command to print the full env config, to be piped to SSH.
```
jkinkead: Seems like you could just use `scp` instead?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -459,92 +486,306 @@ object DeployPlugin extends AutoPlugin {
...
+    val echoConfCommand = Seq("echo", renderedConf)
+    // Shell-friendly version of above, for copy-paste.
+    val quotedEcho = Seq("echo", s"'$renderedConf'").mkString(" ")
+
+    // SSH command to receive `echo` results and redirect them to disk.
+    val writeConfCommand = sshCommand :+ host :+ s"cat > $confPath/env.conf"
```
jkinkead: This method is going to cause problems debugging, since there's no way to generate the `env.conf` file short of running the deploy.

Ideally, generating the `env.conf` file would be a separate SBT task, run before the actual deploy. That way, you could manually generate it (and you could just use `rsync` to push out the new file).

jkinkead:
A couple of comments. The most critical one is the bad `find` pattern for large replica counts.

I think it would also be really nice to have the `env.conf` file show up on the hard drive; although, you'd probably have to generate *all* of the `env.conf` files in one place in the task.

That would actually be pretty useful for debugging . . .

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: :scream: oops

danxmoran: I'm having trouble thinking of a sane solution to this.  Would it be acceptable to just stop every replica matching the pattern `projectName*`, and not delete any directories?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: Should I still attempt to run the `mkdir` anyways? If permission is denied, the error should pop out.

danxmoran:
When I generate the `env.conf` files, should I put them in the staging directory, or in `/tmp`?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: I won't be able to push `env.conf` as part of the existing `rsync` because there needs to be a separate `env.conf` per replica, so I need to create some other directory structure to store all the files and then manually copy the correct one over for each replica (my first attempt used this wonky `echo | ssh cat` approach to avoid dealing with the filesystem). I should be able to use `scp` as you suggested above, though.

danxmoran:
Also, should the SBT task to generate the individual replica `env.conf`s be exposed over the CLI?

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: We quote in a couple of different ways so I don't think that can be wrapped into a clean helper method, but log-exec-fail definitely can.

danxmoran:
This is getting pretty massive, so I'm going to close & refactor into more manageable chunks.

jkinkead:
I don't think I answered the `env.conf` question: Ideally these would go in a sister directory to `stage` (in other words, somewhere in `target`).

Like:
```
target/deploy/hostname1/dirname1/env.conf
target/deploy/hostname2/dirname1/env.conf
```

PR #211, opened 2016-03-02
Remove broken project.version logic.

danxmoran:
As discussed offline, the existing support for `project.version` is broken in the plugin because it checks out the specified git commit AFTER staging, and never re-stages. Since we don't plan to use the functionality for our deploys, it's better to just remove it.

### DIFF
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -36,10 +36,8 @@ import scala.util.{ Failure, Success, Try }
   * override.
   *
   * Overrides of the deploy target's settings can also be specified as arguments to the `deploy`
-  * task via Java property overrides (-Dprop.path=propvalue). This is the easiest way to specify a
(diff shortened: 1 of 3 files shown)


src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -482,9 +470,8 @@ object DeployPlugin extends AutoPlugin {
     * that should be shared between them, when deploying a project.
     * @param hostConfigs configuration objects specifying how the project should be deployed to
     *                    many hosts
-    * @param projectVersion the version of the project being deployed
     */
-  case class DeployConfig(hostConfigs: Seq[HostConfig], projectVersion: Option[String])
+  case class DeployConfig(hostConfigs: Seq[HostConfig])
```
danxmoran: This case class is superfluous within the scope of this PR, but I left it in because I'll add additional fields to it as part of subsequent replica changes.

jkinkead:
Easy one - LGTM!

PR #213, opened 2016-03-02
Move loading of deploy config into task.

danxmoran:
This is in prep for adding a new task which generates env config.


src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -264,30 +264,85 @@ object DeployPlugin extends AutoPlugin {
...
 
-  /** Task used to deploy the current project to a remote host. */
-  lazy val deployTask = deploy := {
-    // Dependencies
-    gitRepoClean.value
+  /** Task used to load deployment configuration for a specific environment from the file
+    * conf/deploy.conf, with optional overrides. Overrides can be specified either as
+    * property-style arguments or as HOCON in ~/.deployrc.
+    *
+    * Takes as arguments:
```
jkinkead: Doesn't this require the user to now type "loadDeployConfigTask prod"? Can you still type "deploy prod"?

Actually, this isn't even a setting key - will it be even visible to the user?

jkinkead:
A couple of questions.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: It's purposefully hidden from the user.  The deploy task calls & uses the return value of this task internally, so as of now this PR is a no-op as compared to having this logic inside a helper function.

I'm moving the logic into a task in anticipation of adding a new `generateEnvConfig` task.  I'd expect that new task to take the same arguments as `deploy` (overrides and target), and as you mentioned on #210 we'll want to write env config to directories named after the remote hosts specified in deploy config.  Both tasks will therefore depend on loading deploy config, and `deploy` will additionally depend on `generateEnvConfig`.  With this dependency graph, the logic for loading deploy config would execute twice every time we run a deploy if it remained in a helper function.  Moving it into a task prevents this thanks to SBT's [evaluation semantics](http://www.scala-sbt.org/0.13/docs/Custom-Settings.html#Execution+semantics+of+tasks) (if their docs are to be believed, at least).

jkinkead: SGTM.

PR #214, opened 2016-03-03
Add SBT task to generate environment configuration files.

danxmoran:
In addition to adding the task and a few related settings, this PR adds helpful tab-completion to the deploy task (as well as the new task). Specifically, typing `deploy <TAB>` will now print out all of the top-level keys listed in `conf/deploy.conf`, as well as a generic pattern for the config overrides we support. Typing `generateEnvConfig <TAB>` will just show the top-level config keys.

After this one the code should be ready to re-introduce the replica logic!


jkinkead:
Awesome, this is starting to look like a real plugin! ;)

LGTM!

PR #215, opened 2016-03-04
Add ability to deploy project to many replicas.

danxmoran:
Finishing up the plugin work.

### DIFF
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -282,13 +282,12 @@ object DeployPlugin extends AutoPlugin {
   /** Wrapper for config common to all hosts in a deploy.
     * @param projectName the name of the current project
     * @param rootDeployDir the root directory to which all replicas should be deployed on every
-    *                      remote host the project is copied to
+    *     remote host the project is copied to
     * @param sshUser the username to use when accessing remote hosts
     * @param startupScript the path, relative to a deployed instance of this project's root, to
-    *                      the script that should be used to restart the project after deploying
-    *                      updates
+    *     the script that should be used to restart the project after deploying updates
     * @param replicasByHost a map from host URLs to collections of info for every replica that
-    *                       should be deployed on the respective hosts
+    *     should be deployed on the respective hosts
     */
   case class DeployConfig(
     projectName: String,
@@ -300,9 +299,9 @@ object DeployPlugin extends AutoPlugin {
 
   /** Config information required to deploy one replica of a project to a remote host.
     * @param directory the directory on the host, relative to the root deploy directory, in which
-    *                  all project code for this replica should be placed
+    *     all project code for this replica should be placed
     * @param envOverrides configuration overrides specific to this replica that should be merged
-    *                     into the application configuration of the deploy environment
+    *     into the application configuration of the deploy environment
     */
   case class ReplicaConfig(
     directory: String,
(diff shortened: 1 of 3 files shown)


src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -391,34 +595,12 @@ object DeployPlugin extends AutoPlugin {
...
-  }
-
-  /** Parser used for reading input for the deploy task. */
-  lazy val deployParser: State => Parser[(Config, String)] = { state =>
-    Space ~> ((overridesParser(state) <~ Space).? ~ deployTargetParser(state)) map {
-      case (overrides, env) => (overrides getOrElse ConfigFactory.empty, env)
-    }
+  /** Wrapper for information used to access a remote host. */
+  class SshInfo(val user: String, val host: String, keyFile: Option[String]) {
+    val keyfileFlag = keyFile map { Seq("-i", _) } getOrElse Seq()
```
jkinkead: You can also call `toIterable` on `keyFile` before calling `map`, instead of using `getOrElse`.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -391,34 +595,12 @@ object DeployPlugin extends AutoPlugin {
...
+    val sshCmd = ("ssh" +: keyfileFlag) ++ Seq("-l", user, host)
```
jkinkead: You should add type declarations to public fields.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -261,7 +277,76 @@ object DeployPlugin extends AutoPlugin {
     stageDir
   }
 
-  /* ==========>  Environment config generation.  <========== */
+  /* ==========>  Deploy config parsing.  <========== */
+
+  /** Wrapper for config common to all hosts in a deploy.
+    * @param projectName the name of the current project
+    * @param rootDeployDir the root directory to which all replicas should be deployed on every
+    *                      remote host the project is copied to
```
jkinkead: I prefer using a four-space indent for continuation lines. It uses less vertical space.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -348,10 +540,22 @@ object DeployPlugin extends AutoPlugin {
...
     }
-    // Write the fully-resolved environment configuration to disk.
-    val targetFile = targetDir / URLEncoder.encode(hostName, "utf-8") / "env.conf"
-    log.info(s"Writing environment config for '$deployEnv' to '${targetFile.getPath}'")
-    IO.write(targetFile, fullEnvConf.root.render(ConfigRenderOptions.concise))
+
+    // Resolve and write config files for all replicas, including their specified overrides.
+    deployConfig.replicasByHost foreach {
+      case (host, replicas) => replicas foreach { replica =>
+        // Encode hostname to reduce chances of naming problems.
```
jkinkead: This is a *tiny* bit worrying. Does it retain `.` literals, or are those encoded? If they are encoded, I'd almost prefer this just fail when you get a weird hostname.

jkinkead:
Just a couple of small comments. It's a little hard to tell what's new and what's just being moved around.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
danxmoran: `.` literals are retained. From the docs:

When encoding a String, the following rules apply:

- The alphanumeric characters "a" through "z", "A" through "Z" and "0" through "9" remain the same.
- The special characters ".", "-", "*", and "_" remain the same.
- The space character "   " is converted into a plus sign "+".
- All other characters are unsafe and are first converted into one or more bytes using some encoding scheme. Then each byte is represented by the 3-character string "%xy", where xy is the two-digit hexadecimal representation of the byte. The recommended encoding scheme to use is UTF-8. However, for compatibility reasons, if an encoding is not specified, then the default encoding of the platform is used.

danxmoran:
Yeah, I think the only "new" functionality is stuff you've already reviewed in #210 before I closed it. I'll address the style comments!

danxmoran:
Comments addressed

jkinkead:
Great, LGTM!

PR #216, opened 2016-03-08
Remove buggy tab-completion from deploy target parser.

danxmoran:
Way more trouble than it's worth.  Fixes calling `deploy` and `generateEnvConfig` in one project from another in a multi-project build.


jkinkead:
Awesome, thanks for the unit tests!

I've struggled with the tab-completion stuff myself in the past. I think you got farther along than I ever managed to. :)

The build is still marked pending here, but it looks like it passed on Semaphore, so I'm merging.

PR #217, opened 2016-03-22
Fix logic for stopping old replicas.

danxmoran:
Ran into an issue trying to test-deploy Contempo updates because Contempo uses `bin/webapp.sh` to start/stop itself but the plugin assumed the script used for stopping would be named `bin/contempo-test.sh`


jkinkead:
LGTM!

PR #218, opened 2016-04-06
Add Jedis to core dependencies.

danxmoran:
Step 1 of centralizing cache usage.


danxmoran:
We use 2.7.2 everywhere I've seen, but the latest version is 2.8.1.  There don't seem to be any super-relevant updates in the changelog.  Worth upgrading anyways?

danxmoran:
Oh wait, nevermind.  This definitely belongs in the common-only dependencies if we're trying to nudge towards universal use of the `JsonQueryCache`.

PR #219, opened 2016-04-07
Make sure deploy only stops what it needs to.

danxmoran:
@OyvindTafjord are you comfortable looking at this, since Jesse's away?  This fixes the issue with the `controller-demo` project.

### DIFF
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -737,9 +737,9 @@ object DeployPlugin extends AutoPlugin {
       rootDeployDirectory, "-maxdepth", "1",
       // Set `find` to only report directory names.
       "-type", "d",
-      // Set `find` to only report directories named after this project (to avoid stopping other
-      // deployed projects).
-      "-name", s""""$projectName*"""",
+      // Set `find` to only report directories named after this project, with an optional numeric
+      // suffix (to avoid stopping other deployed projects).
+      "-regex", s""""./$projectName\(\-[0-9]+\)?"""",
       // Within each replica directory meeting the above criteria, stop the service.
       "-exec", s"{}/$stopScriptPath stop \\;"
     )


src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -737,9 +737,9 @@ object DeployPlugin extends AutoPlugin {
       rootDeployDirectory, "-maxdepth", "1",
       // Set `find` to only report directory names.
       "-type", "d",
-      // Set `find` to only report directories named after this project (to avoid stopping other
-      // deployed projects).
-      "-name", s""""$projectName*"""",
+      // Set `find` to only report directories named after this project, with an optional numeric
+      // suffix (to avoid stopping other deployed projects).
+      "-regex", s""""./$projectName\(\-[0-9]+\)?"""",
```
danxmoran: Context: the deploy plugin supports deploying multiple instances of a single project to a single host, and will place the copies in directories named `$projectName-1`, `$projectName-2`, ... etc. *unless* there's only one copy of the project, in which case the directory is just named `$projectName`.

With this functionality, it's possible that any number of old project instances could be running on a host when we deploy to it, so we need to stop all of those instances before deploying new code.  This `find` command does that, but it was also finding and stopping any other projects with the prefix `$projectName-` (i.e. `controller-demo` for `projectName = controller`).

OyvindTafjord:
Slightly hacky dependence on the format of the project names, but should work as a fix for now, thanks! (btw there's a strange empty "/local/deploy/controller-23" directory on the dev machine right now...?)

danxmoran:
The random directory was me testing that the "find" actually found correct things :)

Agreed this setup is very hacky.  I doubt we're going to bend over backwards to try to support running multiple instances of a project on a single machine when we move to a new deploy tool, since it's consistently more trouble than it's worth.

jkinkead:
I think we want to move to a 1:1 ratio of machines to services in the future. @danxmoran's run into tons of issues with the JVM eating up all extra CPUs for GC. It's just too hard to ensure a timely response from everything when all your resources are contested.

jkinkead:
What I'm saying is: I agree that it doesn't matter that this is a hack, since we probably don't want to be doing this anyway.

PR #220, opened 2016-04-08
Upgrade default common to 1.2.0.

danxmoran:
Pull in the latest version of common, and continue the quest for semi-consistent SBT usage!

### DIFF
--- a/src/main/scala/org/allenai/plugins/CoreDependencies.scala
+++ b/src/main/scala/org/allenai/plugins/CoreDependencies.scala
@@ -79,7 +79,7 @@ trait CoreDependencies {
     * libraryDependencies += Dependencies.allenAiIndexing
     * }}}
     */
-  val allenAiCommonVersion = "1.1.3"
+  val allenAiCommonVersion = "1.2.0"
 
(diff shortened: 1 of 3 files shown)


danxmoran:
@jkinkead after this merges could you cut a release? :)

src/main/scala/org/allenai/plugins/CoreDependencies.scala
```diff
@@ -88,6 +88,14 @@ trait CoreDependencies {
   lazy val allenAiTestkit = allenAiCommonModule("testkit")
   lazy val allenAiWebapp = allenAiCommonModule("webapp")
   lazy val allenAiIndexing = allenAiCommonModule("indexing")
+  lazy val allenAiCache = {
+    val splitVersion = allenAiCommonVersion.split('.')
```
jkinkead: Aside: I hate this method in Scala's enhanced string class, because it looks like it's not using a regex under the covers, but it totally is. Don't use this in critical code sections. :)

jkinkead:
LGTM! I'll cut the release shortly.

PR #221, opened 2016-04-11
Double-escape characters in deploy string to make Scala happy.

danxmoran:
Sigh.

I verified it actually works when run from Scala and not just the command-line this time.


jkinkead:
LGTM!

danxmoran:
Can you cut a release with the fix, please? :)

jkinkead:
Will do!

PR #222, opened 2016-04-13
Fix deploy again.

danxmoran:
@jkinkead [Another regex bug](https://xkcd.com/1171/).  Didn't hit it when testing the last fix because it only applies to deploying an update to an already-running app, and my test deploy last time push Ermine to a fresh machine.


jkinkead:
LGTM!

PR #223, opened 2016-04-19
Create the public node directory before starting a server.

### LINKED ISSUE #169: Webapp Plugin: Front-end build fails on initial build
@ekolve and a few interns have experienced this (possibly non-deterministically) -- but the very first `reStart` on a fresh repository appears to fail to build the front-end into `public`.

jkinkead:
Fixes #169 .

Requires https://github.com/allenai/syrup/pull/59 as well.

### DIFF
--- a/src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
@@ -25,6 +25,8 @@ object WebappPlugin extends AutoPlugin {
     NodeKeys.nodeProjectDir in Npm := (baseDirectory in thisProject).value / "webapp",
     // Run "npm watch" when we run a re-start.
     Revolver.reStart <<= Revolver.reStart.dependsOn(NodeKeys.nwatch in Npm),
+    // Ensure the node target directory is created before the application tries to start up.
+    Revolver.reStart <<= Revolver.reStart.dependsOn(NodeKeys.mkNodeTarget in Npm),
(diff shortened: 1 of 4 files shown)


src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
```diff
@@ -25,6 +25,8 @@ object WebappPlugin extends AutoPlugin {
     NodeKeys.nodeProjectDir in Npm := (baseDirectory in thisProject).value / "webapp",
     // Run "npm watch" when we run a re-start.
     Revolver.reStart <<= Revolver.reStart.dependsOn(NodeKeys.nwatch in Npm),
+    // Ensure the node target directory is created before the application tries to start up.
+    Revolver.reStart <<= Revolver.reStart.dependsOn(NodeKeys.mkNodeTarget in Npm),
```
codeviking: Does the order these dependencies get added matter (i.e. should this happen before starting the `npm run watch`)?

codeviking:
One clarifying question, merge on 🍏

src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
jkinkead: Order doesn't matter, but it's true that this should probably be a dependency on `nwatch` instead. I'll fix it!

jkinkead:
I added some fixes, if you want to take another look.

If I don't hear otherwise, I'll merge on green!

PR #224, opened 2016-04-22
Enable caching in stage command.

danxmoran:
The latest combination of SBT and the native-packager plugin seem to handle this intelligently.

### DIFF
--- a/src/sbt-test/sbt-plugins/cache-key/test
+++ b/src/sbt-test/sbt-plugins/cache-key/test
@@ -16,10 +16,17 @@ $ copy-file webservice/target/universal/stage/conf/cacheKey.Sha1 cacheKey2.Sha1
 > assertCacheKeysEqual
 
 # cache key should change when library dependencies change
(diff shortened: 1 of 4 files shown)


src/sbt-test/sbt-plugins/cache-key/test
```diff
@@ -16,10 +16,17 @@ $ copy-file webservice/target/universal/stage/conf/cacheKey.Sha1 cacheKey2.Sha1
 > assertCacheKeysEqual
 
 # cache key should change when library dependencies change
+-$ exists webservice/target/universal/stage/lib/org.apache.derby.derby-10.4.1.3.jar
```
jkinkead: Can you also add a test that verifies that removing files from [one of the mapped directories](https://github.com/danxmoran/sbt-plugins/blob/2d20916084f23ba85b2e65082f7b7110dbc10496/src/main/scala/org/allenai/plugins/DeployPlugin.scala#L147-L152) is also handled?

jkinkead:
One request-for-test. :)

danxmoran:
Test added!

jkinkead:
Awesome, LGTM!

PR #225, opened 2016-05-05
New deploy task + misc cleanup

jkinkead:
Simple, simple stuff.

LMK if `cleanStage` is used somewhere; I don't think it is.


src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -103,14 +100,17 @@ object DeployPlugin extends AutoPlugin {
 
     val stageAndCacheKey = taskKey[File]("Stage the current project, and calculate its cache key.")
 
+    // Clients may override to perform a different action in addition to staging (i.e. running
+    // database migrations).
+    val stageDeploy =
```
danxmoran: Now wondering about the name...

To me, anything with `stage` in the name shouldn't actually modify a remote system (i.e. migrate a database).  Could we instead have a `preDeploy` task that is empty by default?

Should this also be an input task for the general case?  Or will it get all of its information from environment variables?

danxmoran: (Hoping for environment vars since input tasks are The Worst)

jkinkead: Fair point on naming.

In terms of targets: Initially I might just fudge this and have the e-wk deploy be hard-coded to run the production migration, since there isn't any other deployment of the e-wk.

For a longer-term solution, I think we need to have a notion of the deploy stages being config keys in sbt. In that case, we'd run `$projectName/$stage:preDeploy` or whatever.

danxmoran:
LGTM

PR #226, opened 2016-05-18
(DeployPlugin) Expose task key for deployNpmBuild

markschaake:
This enables projects to stub out the NPM build when staging. We need this control for applications in S2 due to bugs in the npm tooling.

@jkinkead please review

### DIFF
--- a/src/main/scala/org/allenai/plugins/DeployPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DeployPlugin.scala
@@ -103,6 +103,8 @@ object DeployPlugin extends AutoPlugin {
     // Clients may override to perform a different action in addition to staging (i.e. running
     // database migrations).
     val preDeploy = taskKey[Unit]("Prep the project for a deploy. Defaults to `stageAndCacheKey`.")
+
+    val deployNpmBuild = taskKey[Unit]("Runs the npm build")
   }
 
   import autoImport._
@@ -120,6 +122,7 @@ object DeployPlugin extends AutoPlugin {
     generateEnvConfigTask,
     loadDeployConfigTask,
     nodeEnv := "prod",
+    deployNpmBuild := npmBuildTask.value,
     stageAndCacheKeyTask,
     // By default, this only depends on staging and generating the cache key.
     preDeploy := { stageAndCacheKey.value },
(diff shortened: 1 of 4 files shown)


markschaake:
@milescrawford FYI

src/main/scala/org/allenai/plugins/DeployPlugin.scala
```diff
@@ -103,6 +103,8 @@ object DeployPlugin extends AutoPlugin {
     // Clients may override to perform a different action in addition to staging (i.e. running
     // database migrations).
     val preDeploy = taskKey[Unit]("Prep the project for a deploy. Defaults to `stageAndCacheKey`.")
+
+    val deployNpmBuild = taskKey[Unit]("Runs the npm build")
```
jkinkead: I don't love the naming, but I think it helps clarify that this is in the deploy plugin, so it's probably fine to leave as-is.

jkinkead:
LGTM! Merge on green.

src/main/scala/org/allenai/plugins/DeployPlugin.scala
markschaake: Yeah, that is my least favorite part of this PR. I'd prefer to have a config like `npmBuild in Deploy`, but we didn't have that set up already

src/sbt-test/sbt-plugins/simple/build.sbt
```diff
@@ -16,10 +16,18 @@ lazy val core = project.in(file("core")).settings(
 
 lazy val cli = project.in(file("cli")).dependsOn(core).enablePlugins(CliPlugin)
 
+lazy val stubbedDeployNpmBuild = taskKey[Unit]("Verify stubbing the deploy npm build works")
+
 lazy val webService = project.in(file("webservice")).dependsOn(core).enablePlugins(WebServicePlugin)
 
 lazy val webApp = project.in(file("webapp")).dependsOn(core).enablePlugins(WebappPlugin)
-  .settings(NodeKeys.nodeProjectDir in Npm := file("client"))
+  .settings(
```
milescrawford: Want to comment why this is necessary? Or why someone would want to set this?

markschaake: This is a dummy task just for the purpose of testing. The custom task key is defined above in this file. I can walk you through how scripted testing works if you like.

PR #227, opened 2016-06-30
Add AllenAI Bintray publishing locations.

ColinArenz:
Note: This PR contains the release changes so I could test on another repository. I can back those out if desired.

### DIFF
--- a/src/main/scala/org/allenai/plugins/CoreRepositories.scala
+++ b/src/main/scala/org/allenai/plugins/CoreRepositories.scala
@@ -1,9 +1,15 @@
 package org.allenai.plugins
 
+import bintray.BintrayKeys._
 import sbt._
 import sbt.Keys._
 
 object CoreRepositories {
+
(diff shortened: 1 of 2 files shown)


src/main/scala/org/allenai/plugins/CoreRepositories.scala
```diff
@@ -61,5 +67,15 @@ object CoreRepositories {
         if (isSnapshot.value) Some(ai2PublicSnapshots) else Some(ai2PublicReleases)
       }
     }
+
+    val ai2BintrayPrivate = Seq(
```
jkinkead: You may wish to comment these - something like `Settings to make bintray publishing work` or whatever. There's another setting that's needed to enable it as well, right?

ColinArenz: These work in isolation, as long as a license is set.  Projects can add further settings on top of these too.

PR #230, opened 2016-08-03
Remove Nexus and Sonatype resolvers and publish settings.

ColinArenz:
Nexus and Sonatype are deprecated for publishing.  Future updates should all be published to Bintray.  The removed resolvers can be added in long form to projects that still require them.

### DIFF
--- a/src/main/scala/org/allenai/plugins/CoreRepositories.scala
+++ b/src/main/scala/org/allenai/plugins/CoreRepositories.scala
@@ -13,12 +13,8 @@ object CoreRepositories {
 
   /** Common resolvers */
   object Resolvers {
-    private val ai2RepoUrl = "http://utility.allenai.org:8081/nexus/content/repositories"
-
-    val ai2PrivateSnapshots = "AI2 Private Snapshots" at s"${ai2RepoUrl}/snapshots"
-    val ai2PrivateReleases = "AI2 Private Releases" at s"${ai2RepoUrl}/releases"
-    val ai2PublicSnapshots = "AI2 Public Snapshots" at s"${ai2RepoUrl}/public-snapshots"
-    val ai2PublicReleases = "AI2 Public Releases" at s"${ai2RepoUrl}/public-releases"
+    val ai2PrivateReleases = Resolver.bintrayRepo("allenai", "private")
+    val ai2PublicReleases = Resolver.bintrayRepo("allenai", "maven")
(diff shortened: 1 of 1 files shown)


src/main/scala/org/allenai/plugins/CoreRepositories.scala
```diff
@@ -28,8 +24,7 @@ object CoreRepositories {
     val defaults = Seq(
       spray,
```
ColinArenz: I'm skeptical about injecting spray resolvers, but that's for another time.

src/main/scala/org/allenai/plugins/CoreRepositories.scala
```diff
@@ -28,8 +24,7 @@ object CoreRepositories {
...
       Resolver.jcenterRepo,
-      typesafeReleases,
-      ai2PublicReleases
```
ColinArenz: I can leave this in (which auto-adds Bintray), but I think it might be better to just require projects to do this explicitly.

schmmd: I'm wary of removing `typesafeReleases`--mostly because I don't know what impact it will have.  I'd hate for everyone in every project to have to remember to add that back in.

schmmd:
LGTM, but I'm assigning over to Jesse since he's much more familiar with this codebase!

src/main/scala/org/allenai/plugins/CoreRepositories.scala
ColinArenz: That one is still there.  It shows up as a remove here because the trailing comma was removed on this line.

schmmd: Doh! Thanks for pointing that out, Colin.

src/main/scala/org/allenai/plugins/CoreRepositories.scala
jkinkead: Yeah, I can see taking this out.

jkinkead:
LGTM! Make sure you do a major version bump & good release notes for this. It could cause strange breakage in consumers.

jkinkead:
Also, be sure you run the plugin tests!

ColinArenz:
Tests pass, merging.  I'll update versions in a separate change.

PR #232, opened 2016-10-25
Pull a couple of methods needed for Docker into a Utilities class.

jkinkead:
Follow best-practices with respect to resource location.

This is a very simple PR. Note the sbt tests, which are deeply weird.


src/main/scala/org/allenai/plugins/Utilities.scala
```diff
@@ -0,0 +1,52 @@
...
+
+    jarName.append(".jar").toString
+  }
+
+  /** Copies a class resource with the given name to a given location.
+    * @param clazz the class to load the resource from
+    * @param resourceName the name of the resource to copy
+    * @param destination the destination file for the resource
+    */
+  def copyResourceToFile(clazz: Class[_], resourceName: String, destination: File): Unit = {
```
ColinArenz: Recommend `clazz` -> `classToLoad` or similar.

jkinkead: This is a Java-land idiom. It's used essentially everywhere there's a `class` variable.

The accepted answer here agrees with me:
http://stackoverflow.com/questions/2529974/why-do-java-programmers-like-to-name-a-variable-clazz

ColinArenz: Ah, makes sense.  Thanks for the link.

PR #233, opened 2016-10-26
Add all of the keys needed to configure the new Docker plugin.

jkinkead:
Add default values for setting keys.

Some sbt / plugin notes:

Classes that extend `AutoPlugin` are put in your `.sbt` file's scope automatically. This is how the test `build.sbt` file can refer to `DockerBuildFile`. Additionally, all of the values in any `AutoPlugin`'s `autoImport` field are imported into the `.sbt` file's scope.

sbt keys (like `TaskKey` and `SettingKey`) are values that you can refer to within the sbt  console - things like `re-start` and `compile`. You can apply`Def.Initialize` blocks to them via the `:=` operator. Note that this doesn't actually *mutate* the key - it's a function that returns a `Def.Setting`. This is why you'll see code like:
```scala
lazy val foo = project.in("foo").settings(bar := "yay", gaz := "boo")
```
Any statement in a `.sbt` file whose result is of type `Def.Setting` will have that setting applied to the current project. This is how you actually make changes to a project.

When you `enable` a plugin on a project, all that happens is that the `projectSettings` method is called on it, and the returned settings are included in the project's settings immediately.


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -0,0 +1,135 @@
...
+}
+import sbt.plugins.JvmPlugin
+
+import java.io.File
+
+/** Plugin for building docker images. */
+object DockerBuildPlugin extends AutoPlugin {
+  val AI2_PRIVATE_REGISTRY = "allenai-docker-private-docker.bintray.io"
+
+  val DEFAULT_BASE_IMAGE = AI2_PRIVATE_REGISTRY + "/java:8"
```
ColinArenz: It seems like we should be using a different image name than the public one, if we're customizing it.  If we're not, we should probably pull this from the public registry.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  object autoImport {
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // The following settings affect both the images and Dockerfiles generated by this plugin. When
+    // you update these settings in a build.sbt file, you'll want to re-generate your Dockerfiles.
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+
+    val dockerfileLocation: SettingKey[File] = settingKey[File](
+      "The location of the Dockerfile to use in building the main project image. Defaults to " +
+        "`srcDirectory.value + \"docker/Dockerfile\", typically \"src/main/docker/Dockerfile\"."
```
ColinArenz: Missing backtick.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -0,0 +1,135 @@
...
+    // image tags will be, for the main image:
+    //   ${imageRegistryHost}/${imageNamePrefix}/${imageName}
+    // and for the dependency image will be:
+    //   ${imageRegistryHost}/${imageNamePrefix}/${imageName}-dependency
+    //
+    // See the documentation for details on which tags will be used by `dockerBuild` and
+    // `dockerPush`.
+    val imageRegistryHost: SettingKey[String] = settingKey[String](
+      "The base name of the image you're creating. Defaults to " +
+        "allenai-docker-private-docker.bintray.io ."
```
ColinArenz: Use `AI2_PRIVATE_REGISTRY` instead of hard-coding if possible.  Even better, can you describe these defaults by referencing `projectSettings` below?

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -0,0 +1,135 @@
...
+    )
+    val imageNamePrefix: SettingKey[String] = settingKey[String](
+      "The image name prefix (\"repository\", in Docker terms) of the image you're creating. " +
+        "Defaults to organization.value.stripPrefix(\"org.allenai.\") . " +
+        "This is typically the github repository name."
+    )
+    val imageName: SettingKey[String] = settingKey[String](
```
ColinArenz: Will this be prefixed with something when used?  `imageName` doesn't scream "Docker plugin" to me, similar feedback for other names.  Perhaps they should all be prefixed `docker` like `dockerWorkdir` below?

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
jkinkead: This is nominally customized (really, just in that it uses a specific ubuntu version).

Since the registry is part of the image name, it seemed fine to me to use `java` directly. I mean, we could do `allenai/java`, which I initially had, but that seems weird with the registry prefix.

There's nothing stopping us from making this be `allenai/java` if it were ever to be released into the wild on Docker hub.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
jkinkead: Fixed.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
ColinArenz: True.  I think the main thing I'm looking for is that if I forget to add the registry I get a failure instead of pulling the wrong image.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
jkinkead: Constant used.

These are online help strings used by `inspect` in the sbt console (and possibly with other sbt commands?), so they should contain the defaults rather than point to code.

Documenting defaults at all is a little awkward, since the keys *technically* don't have a default until you enable the plugin. I think it's a good balance to have them documented here.

Anyone reading the plugin code will hopefully know enough to look at the `projectSettings` method.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
jkinkead: Hmm, interesting. I'm not sure I think that it matters w.r.t. a plugin, but it probably matters for hand-authored files.

jkinkead:
PR comments addressed - TAL?

jkinkead:
Thanks!

PR #234, opened 2016-10-27
Add a file hashing utility to the Utilities class.

jkinkead:
This makes cache key generation no longer sensitive to file ordering, but adds sensitivity to file naming.

One more utility method for generating file hashes. This will be used a few times in the Docker image generation code.

### DIFF
--- a/src/main/scala/org/allenai/plugins/Utilities.scala
+++ b/src/main/scala/org/allenai/plugins/Utilities.scala
@@ -59,8 +59,9 @@ object Utilities {
     */
   def hashFiles(filesToHash: Seq[File], rootDir: File): String = {
     // Resolve the filenames relative to the root directory.
-    val rootDirPath = rootDir.toPath
-    val relativizedNames = filesToHash.map(_.toPath).map(rootDirPath.relativize).map(_.toString)
+    val rootDirPath = rootDir.toPath.normalize
+    val relativizedNames =
+      filesToHash.map(_.toPath.normalize).map(rootDirPath.relativize).map(_.toString)
     // Create a hash of the sorted names, joined by an empty string.
     val nameHash = Hash.toHex(Hash(relativizedNames.sorted.mkString))
 
--- a/src/main/scala/org/allenai/plugins/Utilities.scala
+++ b/src/main/scala/org/allenai/plugins/Utilities.scala
@@ -1,6 +1,6 @@
 package org.allenai.plugins
 
-import sbt.{ Artifact, IO, ModuleID }
+import sbt.{ Artifact, Hash, IO, ModuleID }
 
 import java.io.File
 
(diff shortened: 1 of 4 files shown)


src/main/scala/org/allenai/plugins/Utilities.scala
```diff
@@ -49,4 +49,25 @@ object Utilities {
...
+    * change when the contents of the files change, or when the file name changes relative to the
+    * given file.
+    * @param filesToHash the files whose contents and names should be hashed
+    * @param rootDir the root directory the file names should be resolved against before hashing.
+    * Only path changes relative to this directory will cause the hash to change.
+    */
+  def hashFiles(filesToHash: Seq[File], rootDir: File): String = {
+    // Resolve the filenames relative to the root directory.
+    val rootDirPath = rootDir.toPath
+    val relativizedNames = filesToHash.map(_.toPath).map(rootDirPath.relativize).map(_.toString)
```
ColinArenz: I assume relativize will create a canonical path?  If I feed it "foo/dir/../dir" what happens?  Probably add a test for this.

jkinkead: It does, yes.

Test added!

jkinkead: Oh, you were asking another thing! I'll check that out.

jkinkead: Ah-ha! It does not. Good catch; I'll add a test.

jkinkead: Another test added. :)

I'll merge on green, unless I hear otherwise!

jkinkead:
Actually, there is no semaphore integration here, so I will just merge one I run the tests again locally!

src/test/scala/org/allenai/plugins/UtilitiesSpec.scala
```diff
@@ -76,6 +76,19 @@ class UtilitiesSpec extends FlatSpecLike with Matchers with OneInstancePerTest w
...
 
+  it should "return the same hash when files have the same normalized path" in {
+    // Create versions of fooFile and barFile which are semantically the same, but have different
+    // paths.
+    val fooRelative = new File("foo.txt")
+    val barRelative = new File(tempDirectory, ".." + File.pathSeparatorChar +
+      tempDirectory.getName + File.pathSeparatorChar + "bar.txt")
+
+    val firstHash = Utilities.hashFiles(Seq(fooFile, barFile), tempDirectory)
+    val secondHash = Utilities.hashFiles(Seq(fooFile, barFile), tempDirectory)
```
ColinArenz: Shouldn't the second hash attempt use a different (denormalized) path to barFile?

jkinkead: Hah! Indeed. I'll send a follow-up.

PR #236, opened 2016-10-31
Docker stage commands.

jkinkead:
This also includes some scripted test refactoring.

Note that there is a reference to the `generateDockerfile` task, which doesn't have an implementation yet. That's next, along with actual docker interaction (build, push, run).


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -112,23 +117,175 @@ object DockerBuildPlugin extends AutoPlugin {
...
+    // Create the destination directory.
+    val imageDirectory = dependencyImageDir.value
+    IO.createDirectory(imageDirectory)
+    val lib = dependencyLibDir.value
+    IO.createDirectory(lib)
+
+    // Create the Dockerfile for the dependency image.
+    val dockerfileContents = s"""
+      |FROM ${dockerImageBase.value}
+      |
```
ColinArenz: Nit-pick: Recommend removing the empty-line padding, since this file is so short.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -112,23 +117,175 @@ object DockerBuildPlugin extends AutoPlugin {
...
+      |WORKDIR /local/deploy/${dockerWorkdir.value}
```
ColinArenz: I'd remove `/local/deploy/` here and make `dockerWorkdir` an absolute path.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -112,23 +117,175 @@ object DockerBuildPlugin extends AutoPlugin {
...
+    dockerfileLocation := {
+      new File(sourceMain.value, "docker" + File.separatorChar + "Dockerfile")
+    },
     dockerCopyMappings := defaultCopyMappings.value,
     dockerImageRegistryHost := AI2_PRIVATE_REGISTRY,
     dockerImageNamePrefix := Keys.organization.value.stripPrefix("org.allenai."),
     dockerImageName := Keys.name.value,
     dockerImageBase := DEFAULT_BASE_IMAGE,
-    dockerWorkdir := "/local/deploy/" + dockerImageName.value
+    dockerWorkdir := dockerImageName.value,
```
ColinArenz: I think this default should be simply `"/stage"`, not image dependent.  We can leave the override in case someone wants to combine images in unexpected ways, but I don't want to encourage that pattern up front.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -112,23 +117,175 @@ object DockerBuildPlugin extends AutoPlugin {
...
-    // See http://www.scala-sbt.org/0.12.3/docs/Detailed-Topics/Mapping-Files.html
-    // for more info on sbt mappings.
-    Seq("bin", "conf").map { dir =>
-      PathFinder(sourceMain.resolve(dir).toFile).***.pair(relativeTo(sourceMain.toFile))
-    }.flatten
+    Seq("bin", "conf").map { dir => (new File(dir), dir) }
+  }
+
+  /** The full image name, derived from the user-provided settings. */
+  lazy val fullImageName: Def.Initialize[String] = Def.setting {
```
ColinArenz: Do we want a `docker` prefix on this and `dependencyImageName` below?

jkinkead: These are not setting keys - they can't be overridden - and they aren't in `autoImport`, so they won't show up in the `sbt` console. If someone wanted to use them in Scala code, they could do `DockerBuildPlugin.fullImageName`, which I think is qualified-enough without needing the extra `docker` prefix.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
jkinkead: Done.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
jkinkead: Sure, SGTM. Fixed.

jkinkead:
Comments addressed!

PR #239, opened 2016-11-03
Fix most deprecation warnings introduced in the upgrade to 0.13.13.

jkinkead:
Ugh. This turned out to be a bit of a nightmare, since two of the warnings seem unfixable with current sbt technology.

Anyway, I confirmed these all pass the current test suite, and seem to work with Aristo.


jkinkead:
Actually, scratch this - I have a better way to do these.

PR #240, opened 2016-11-03
Fix deprecation warnings introduced in the upgrade to 0.13.13.

jkinkead:
I figured out how to fix the two last warnings, AND figured out the magic concise `dependsOn` invocation. Hooray!

### DIFF
--- a/src/main/scala/org/allenai/plugins/StylePlugin.scala
+++ b/src/main/scala/org/allenai/plugins/StylePlugin.scala
@@ -37,11 +37,13 @@ object StylePlugin extends AutoPlugin {
       inConfig(Compile)(configSettings) ++
       inConfig(Test)(configSettings) ++
       Seq(
-        // Check style on compile.
-        compileInputs in (Test, compile) <<=
-          (compileInputs in (Test, compile)) dependsOn (StyleKeys.styleCheck in Test),
-        compileInputs in (Compile, compile) <<=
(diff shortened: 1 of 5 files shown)


jkinkead:
Oh, it's worth noting that this also has a couple of formatting fixes.

src/main/scala/org/allenai/plugins/StylePlugin.scala
```diff
@@ -100,22 +102,20 @@ class SbtLogOutput[T <: FileSpec](logger: Logger) extends ScalastyleOutput[T] {
...
   }
 
-  private[this]
-  def location(file: T, line: Option[Int], column: Option[Int]): String =
+  private[this] def location(file: T, line: Option[Int], column: Option[Int]): String =
     (file.name +
-     line.map(n => ":" + n + column.map(":" + _).getOrElse(""))
-         .getOrElse(""))
+      line.map(n => ":" + n + column.map(":" + _).getOrElse(""))
+      .getOrElse(""))
```
ColinArenz: Nit-pick: Slightly weird formatting here is hard to follow. Maybe a block with some intermediate variables?

jkinkead: IIRC much of this plugin was copy-pasted from another sbt plugin's implementation.

I made this use a procedural `StringBuilder` because I am a REBEL.

ColinArenz: Living dangerously. I like the new version much better.

PR #244, opened 2016-11-04
Use a path relative to `baseDirectory` for docker mappings.

### LINKED ISSUE #242: `dockerCopyMappings` output should be relative.
When issuing an absolute mapping path, the output should be converted to a path relative to the source directory.  Steps to reproduce below...

1. In `build.sbt`, add this line:
```sbt
dockerCopyMappings += ((file(baseDirectory.value + "/data"), "data"))
```

2. On the command line:
```bash
sbt generateDockerfile
```

Resulting output:
```dockerfile
# The directories in the staging directory which will be mapping into the Docker image.
#   dockerCopyMappings := Seq(
#     (file("resources"), "conf"),
#     (file("/Users/colinarenz/git/<project>/data"), "data")
#   )
COPY conf conf
COPY data data
```

A couple possible solutions jump to mind:
- Forbid absolute paths (make it an error).
- Convert paths to project-relative, rather passing through directly.

It also seems like the first tuple element in dockerCopyMappings should be project-relative rather than relative to `src/main`.

jkinkead:
Print out a relative path in the helper comment in the generated
Dockerfile.

Fixes #242 .

### DIFF
--- a/src/sbt-test/sbt-plugins/simple/docker/build.sbt
+++ b/src/sbt-test/sbt-plugins/simple/docker/build.sbt
@@ -1,8 +1,8 @@
 // Map `foo` in the project root to `bar` in the docker image.
-dockerCopyMappings += (baseDirectory.value / "foo", "bar")
+dockerCopyMappings += (file("foo"), "bar")
 
 // Map src/main/sourcey to `sourcey` in the docker image.
-dockerCopyMappings += (file("sourcey"), "sourcey")
+dockerCopyMappings += (sourceDirectory.value / "main" / "sourcey", "sourcey")
 
 lazy val appendToDockerfile = taskKey[Unit]("append to dockerfile")
 appendToDockerfile := {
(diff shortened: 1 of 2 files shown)


src/sbt-test/sbt-plugins/simple/docker/build.sbt
```diff
@@ -1,8 +1,8 @@
 // Map `foo` in the project root to `bar` in the docker image.
-dockerCopyMappings += (baseDirectory.value / "foo", "bar")
+dockerCopyMappings += (file("foo"), "bar")
 
 // Map src/main/sourcey to `sourcey` in the docker image.
-dockerCopyMappings += (file("sourcey"), "sourcey")
+dockerCopyMappings += (sourceDirectory.value / "main" / "sourcey", "sourcey")
```
ColinArenz: You use `sourceMain.value` elsewhere. Would that be more appropriate here?

jkinkead: That's not part of the public API. This is in integration test of the public API.

PR #246, opened 2016-11-04
Docker build implementation

jkinkead:
A complete build implementation. I added a `dockerDependencyBuild` because it's nice-to-have for debugging.

Note that I split this into three commits; the first two commits are tiny & might be slightly distracting from the main bit of work.

To get the new image tag to print, you can use the `show` command in sbt:
```
> show dockerBuild
[lots and lots of output]
[info] allenai-docker-private-docker.bintray.io/project/subproject:dc1b889c84ce0960784b0459d6db097e512b060c
[success] Total time: 42 s, completed Nov 4, 2016 3:05:38 PM
>
```

### DIFF
--- a/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
@@ -302,7 +302,7 @@ $DOCKERFILE_SIGIL
       if (remainder.nonEmpty) {
         remainder.tail.mkString("\n")
       } else {
-        logger.warn(s"Overwriting Dockerfile at $dockerfile . . .")
+        logger.warn(s"Overwriting Dockerfile at $dockerfile...")
         ""
       }
     } else {
--- a/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
@@ -179,25 +179,6 @@ object DockerBuildPlugin extends AutoPlugin {
     new File(dockerTargetDir.value, "main")
   }
 
-  /** The location of the staged dependency image's library directory, containing all staged jars.
-    */
-  lazy val dependencyLibDir: Def.Initialize[File] = Def.setting {
-    new File(dependencyImageDir.value, "lib")
-  }
-
-  /** The location of the staged dependency image's Dockerfile. */
-  lazy val dependencyDockerfile: Def.Initialize[File] = Def.setting {
-    new File(dependencyImageDir.value, "Dockerfile")
-  }
-
-  /** The location of the staged dependency image's bin directory, containing the startup script. */
-  lazy val dependencyImageBinDir: Def.Initialize[File] = Def.setting {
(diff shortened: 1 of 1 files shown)


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -426,6 +510,26 @@ $DOCKERFILE_SIGIL
...
+  lazy val mainImageBuildDef: Def.Initialize[Task[String]] = Def.task {
+    // This task requires docker to be installed.
+    requireDocker.value
+
+    // This task requires that the dependency image be created, and that the main image be staged.
+    mainImageStageDef.value
+    dependencyBuildDef.value
+
+    val logger = Keys.streams.value.log
+    logger.info("Building main image...")
```
ColinArenz: I've been noticing that when running `sbt generateDockerfile` globally, these generic logging messages have been really hard to localize to a specific project.  It looks like that's the case here too.  Can you add a line signaling what's project is currently building?

Note: Separate change is fine, but this may be an opportunity since you're in here anyway.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -156,9 +164,59 @@ object DockerBuildPlugin extends AutoPlugin {
...
+      IO.read(hashFile)
+    } else {
+      ""
+    }
+
+    val imageNameWithLabel = imageName + ':' + newHash
+    if (newHash != oldHash) {
+      // Build a new docker image.
+      val buildCommand =
+        Seq("docker", "build", "-t", imageName, "-t", imageNameWithLabel, imageDir.toString)
```
ColinArenz: Does the naked image tag (`:latest` ) get pushed to the registry when pushing the SHA-tagged one, or do they have to be pushed separately?  I'm hoping `:latest` is a local-only tag.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -179,6 +237,11 @@ object DockerBuildPlugin extends AutoPlugin {
     new File(dockerTargetDir.value, "main")
   }
 
+  /** Task which requires that `docker` exists on the commandline path. */
+  lazy val requireDocker: Def.Initialize[Task[Unit]] = Def.task {
+    if (Process(Seq("which", "docker")).!(Utilities.NIL_PROCESS_LOGGER) != 0) {
+      sys.error("`docker` not found on path. Please install the docker client before using this.")
```
ColinArenz: Nit-pick: `s/ before using this//`

jkinkead: Fixed.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
jkinkead: There might be a way to make this work simply using sbt's builtins. I'll play around a little tomorrow & send you another PR once I get it working.

PR #249, opened 2016-11-07
Expose the fully resolved image tags generated by the Docker plugin.

ColinArenz:
These tags will be needed by the `dockerPush` command and also for scripted deployment.  Reconstructing the tags given only a project name would be a bit messy and error-prone outside of SBT.

### DIFF
--- a/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
@@ -138,6 +138,18 @@ object DockerBuildPlugin extends AutoPlugin {
     val dockerKill: TaskKey[Unit] = taskKey[Unit](
       "Kills any currently-running docker container for this project."
     )
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // The following keys expose information about the docker file being generated.
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+
+    val dockerDependencyImageTag: TaskKey[String] = taskKey[String](
+      "Generates and prints the tag of the dependency image that `dockerBuild` will produce."
+    )
+
+    val dockerMainImageTag: TaskKey[String] = taskKey[String](
+      "Generates and prints the tag of the main image that `dockerBuild` will produce."
+    )
   }
   import autoImport._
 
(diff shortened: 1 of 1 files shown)


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -534,6 +553,20 @@ $DOCKERFILE_SIGIL
...
+    val imageTag = imageNameWithLabel(dependencyImageDir.value, dependencyImageName.value)
+    val logger = Keys.streams.value.log
+    logger.info(imageTag)
+    imageTag
+  }
+
+  lazy val mainImageTagDef: Def.Initialize[Task[String]] = Def.task {
+    val imageTag = imageNameWithLabel(mainImageDir.value, mainImageName.value)
+    val logger = Keys.streams.value.log
+    logger.info(imageTag)
```
jkinkead: If you use the `show` command in the sbt console, you don't need this log. It literally does exactly this.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -138,6 +138,18 @@ object DockerBuildPlugin extends AutoPlugin {
     val dockerKill: TaskKey[Unit] = taskKey[Unit](
       "Kills any currently-running docker container for this project."
     )
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // The following keys expose information about the docker file being generated.
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+
+    val dockerDependencyImageTag: TaskKey[String] = taskKey[String](
+      "Generates and prints the tag of the dependency image that `dockerBuild` will produce."
```
ColinArenz: I'm not actually sure... I assumed I had to push both.  If not, I'll remove it.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -171,6 +183,15 @@ object DockerBuildPlugin extends AutoPlugin {
   // Common settings / tasks / utilities used across tasks.
   //////////////////////////////////////////////////////////////////////////////////////////////////
 
+  def hashDirectory(imageDir: File): String = {
+    val allFiles = PathFinder(imageDir).***.filter(_.isFile).get
+    Utilities.hashFiles(allFiles, imageDir)
+  }
+
+  def imageNameWithLabel(imageDir: File, imageName: String): String = {
```
ColinArenz: Removed the directory hashing so it has to be called explicitly each time. This reduces the function to a simple string concatenation, but I think we should keep it for expressiveness.

ColinArenz:
From offline: There are a few other ways to solve this problem that don't involve this plugin.  I'm going to retract this change, at least for now.

PR #254, opened 2016-11-08
Add dockerRun and dockerStop commands.

jkinkead:
Add a dockerPortMapping setting to customize ports mapped by dockerRun.

### DIFF
--- a/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
@@ -252,15 +252,15 @@ object DockerBuildPlugin extends AutoPlugin {
   }
 
   /** Stops the given docker container. This will first rename the container to a UUID, then run
-    * `docker stop` against the new name. Only the rename will block. The rename is done to avoid
-    * race conditions in the docker daemon: It doesn't clean up the name of a stopped container
-    * before `docker stop` exits, so immediately restarting the container results in an error.
+    * `docker stop` against the new name. The rename is done to avoid race conditions in the docker
+    * daemon: It doesn't clean up the name of a stopped container before `docker stop` exits, so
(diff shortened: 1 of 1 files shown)


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -223,6 +251,18 @@ object DockerBuildPlugin extends AutoPlugin {
...
+  /** Stops the given docker container. This will first rename the container to a UUID, then run
+    * `docker stop` against the new name. Only the rename will block. The rename is done to avoid
+    * race conditions in the docker daemon: It doesn't clean up the name of a stopped container
+    * before `docker stop` exits, so immediately restarting the container results in an error.
+    */
+  def stopContainer(container: String, logger: Logger): Unit = {
+    logger.info(s"Stopping $container...")
+    val newName = container + '-' + UUID.randomUUID.toString
+    Process(Seq("docker", "rename", container, newName)).!(Utilities.NIL_PROCESS_LOGGER)
+    Process(Seq("docker", "stop", newName)).run(Utilities.NIL_PROCESS_LOGGER)
```
mjlangan: Just curious, are there ways these commands could fail that we might need to address?

jkinkead: Probably not. If the rename fails, the next `dockerRun` will fail with a helpful message. If the stop fails, the port binding for the next `dockerRun` will fail.

Hopefully this doesn't happen often. :)

jkinkead: Actually, I think I'm going to have the `stop` command also block, since I want that port binding race condition to not happen.

jkinkead:
I'll plan on merging after this build completes, unless I hear otherwise!

PR #258, opened 2016-11-10
Don't call `in` function infix-style.

jkinkead:
Style cleanup. For whichever of you has a spare few minutes.


OyvindTafjord:
LGTM!

PR #262, opened 2016-11-18
Add a verifier to check if generated Dockerfiles are up-to-date.

jkinkead:
This is slightly more complicated than I'd like, due to some code duplication. Due to how sbt handles task dependencies, I don't think there's a cleaner way.


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -129,6 +129,20 @@ object DockerBuildPlugin extends AutoPlugin {
...
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+
+    val verifyDockerfileIsStrict: SettingKey[Boolean] = Def.settingKey[Boolean](
+      "If true, the `verifyDockerfile` task will raise an error if the current Dockerfile does " +
+        "not match the generated one. Defaults to true."
+    )
+
+    val verifyDockerfileOnBuild: SettingKey[Boolean] = Def.settingKey[Boolean](
+      "If true, `dockerBuild` will depend on `verifyDockerfile`, and will not build if the " +
+        "Dockerfile is not up-to-date. Defaults to false."
```
ColinArenz: Why default to false?  It seems like this is generally the desired behavior, given how easy it is to fall out of sync.

jkinkead: I think it's more correct to have this as a CI check, not a build-time check. I think it'll cause a lot of annoyances if it's happening on all builds.

Having a warning on all builds by default is good, but I was getting lazy and didn't want to add a third option . . .

ColinArenz: That makes sense. I think the warning would be a nice-to-have, but I wouldn't block this on it.  👍

jkinkead: OK! I'm adding, it's fairly easy.

jkinkead: I take that back - it's easy, but not very useful, since the warning gets swallowed in the rest of the build output.

I think forcing this to error is probably a better solution.

PR #264, opened 2016-11-22
Make the docker plugin work out-of-the-box with WebappPlugin and Node

jkinkead:
FYI @codeviking , @mjlangan .

### DIFF
--- a/src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
@@ -59,7 +59,7 @@ object WebappPlugin extends AutoPlugin {
       val volumeMapping = {
         val nodeDir = nodeTargetDef.value
         val imageTarget = DockerBuildPlugin.autoImport.dockerWorkdir.value + '/' + nodeDir.getName
-        s"$nodeDir:$imageTarget"
+        s"$nodeDir:$imageTarget:ro"
       }
       Seq("-v", volumeMapping)
     }
--- a/src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
@@ -1,23 +1,20 @@
 package org.allenai.plugins.archetypes
 
-import org.allenai.plugins.NodeJsPlugin
-import org.allenai.plugins.NodeJsPlugin.autoImport._
-import org.allenai.plugins.DeployPlugin
+import org.allenai.plugins.{ DeployPlugin, NodeJsPlugin }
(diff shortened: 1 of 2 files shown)


src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
```diff
@@ -1,38 +1,67 @@
...
+      DockerBuildPlugin.autoImport.dockerMainStage.dependsOn(
+        DeployPlugin.autoImport.deployNpmBuild
+      ).value
+    },
+    // Map the node target directory into the docker image.
+    DockerBuildPlugin.autoImport.dockerCopyMappings ++=
+      MappingsHelper.directory(nodeTargetDef.value),
+    // Make `dockerRun` execute node watch, and `dockerStop` stop the watch.
+    DockerBuildPlugin.autoImport.dockerRun :=
+      DockerBuildPlugin.autoImport.dockerRun.dependsOn(NodeKeys.nwatch.in(Npm)).value,
```
ColinArenz: I'm a bit concerned here about a rogue watcher. What happens if SBT is killed/interrupted or the container is stopped via another shell process?  Will the watch still get stopped at some point?  If not, does the next `dockerRun` overwrite it?

src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
```diff
@@ -1,38 +1,67 @@
...
+    DockerBuildPlugin.autoImport.dockerStop :=
+      DockerBuildPlugin.autoImport.dockerStop.dependsOn(NodeKeys.unwatch.in(Npm)).value,
+    // When running with `dockerRun`, map the local filesystem's target directory in directly so
+    // that rebuilds don't need to happen to see client-side changes.
+    DockerBuildPlugin.autoImport.dockerRunFlags ++= {
+      val volumeMapping = {
+        val nodeDir = nodeTargetDef.value
+        val imageTarget = DockerBuildPlugin.autoImport.dockerWorkdir.value + '/' + nodeDir.getName
+        s"$nodeDir:$imageTarget"
```
ColinArenz: Add the `:readonly` (or `:ro`) modifier, assuming the container doesn't need to modify the JS source.

codeviking:
I don't think I have anything to add, as it's safe to say we won't be using this on S2.  LMK if that's concerning (I'm not sure if you intended for us to eventually port to this)?

jkinkead:
I'll follow-up offline, @codeviking .

src/main/scala/org/allenai/plugins/archetypes/WebappPlugin.scala
jkinkead: Done.

Issue #268, opened 2016-12-10
dockerBuild fails after running 'docker rmi'


ColinArenz:
To reproduce:
- Build a docker-enabled project via `dockerBuild`
- Run `docker rmi <dependency_tag>`
- Run `dockerBuild` again.

The build will fail because `dockerBuild` doesn't rebuild the dependency image if the hash doesn't change.  This causes Docker to issue an error `Error: image <dependency_tag> not found.`.  Note that this will also happen for the main image if issuing a `dockerPush` comand.

I believe the best fix for this would be to re-run `dockerBuild` if `docker inspect --type image <tag>` returns a non-zero exit code.

PR #270, opened 2016-12-12
Misc bugfixes for the docker plugin

### LINKED ISSUE #266: During dockerBuild, dockerCopyMappings doesn't retain file flags
The ENTRYPOINT directive requires its target to be executable.  If this target is a script copied through `dockerCopyMappings`, the file doesn't retain its +x flag.  Though the build succeeds, this ultimately causes the container to fail to start.

The workaround from Jesse works:
```dockerfile
ENTRYPOINT ["sh", "-c", "chmod +x my_script; exec my_script"]
```

jkinkead:
This fixes #266, #268, and a few cache key bugs.

### DIFF
--- a/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
@@ -256,7 +256,14 @@ object DockerBuildPlugin extends AutoPlugin {
     }
 
     val imageNameWithLabel = imageName + ':' + newHash
-    if (newHash != oldHash) {
+
+    // Invalidate the hash if the image isn't saved, for any reason.
+    val imageMissing = {
+      val inspect = Seq("docker", "inspect", "--type", "image", imageName, imageNameWithLabel)
+      Process(inspect).!(Utilities.NIL_PROCESS_LOGGER) != 0
+    }
+
+    if (newHash != oldHash || imageMissing) {
       // Build a new docker image.
       val buildCommand =
         Seq("docker", "build", "-t", imageName, "-t", imageNameWithLabel, imageDir.toString)
--- a/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
@@ -10,6 +10,7 @@ import sbt.{
   InputTask,
   Keys,
   Logger,
+  Path,
   PathFinder,
   Plugins,
   SettingKey,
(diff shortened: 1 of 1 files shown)


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -625,11 +633,23 @@ $DOCKERFILE_SIGIL
...
           if (source.isDirectory) {
             IO.createDirectory(destination)
-            IO.copyDirectory(source, destination)
+            val toCopy = PathFinder(source).***.pair(Path.rebase(source, destination))
+            IO.copy(toCopy)
+            toCopy.foreach {
+              case (source, destination) =>
+                if (source.canExecute) {
+                  destination.setExecutable(true)
+                }
```
ColinArenz: I'd replace this `if` block with `destination.setExecutable(source.canExecute)`.  That way if we'll sync up the flag if the destination was marked executable, but the source's flag has changed.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -649,29 +669,34 @@ $DOCKERFILE_SIGIL
...
       }
 
-      // Find the git commits / random IDs for the current project and local dependencies.
-      val unstableContentsCommits = {
-        Def.taskDyn(gitCommitOrGuid.all(HelperDefs.dependencyFilter.value)).value :+
-          gitCommitOrGuid.value
+      // Find the git commits for the current project and local dependencies, since this is more
+      // stable than a hash of the contents.
+      val unstableContentsCommits: Seq[Option[String]] = {
+        (Def.taskDyn(gitCommitIfClean.all(HelperDefs.dependencyFilter.value)).value :+
```
ColinArenz: I really like that you're doing this by project, so an unrelated change won't dirty it.

jkinkead: There are some edge-cases where this breaks, unfortunately. Notably, this doesn't take into account changes in root sbt settings (settings in the root `build.sbt` or root `project/*`). However, these would have to be changes which affected the compiled code but not either dependencies or the Dockerfile - so, things like updates to the docker plugin...

:P

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
jkinkead: I think the `copy` line beforehand will replace the target, so this might work regardless - but this is cleaner. Updated!

PR #271, opened 2017-01-09
Fix git issues

jkinkead:
I'm not sure if you've run into this, but you can't use our plugins at all if you're not in a git repository.

This fixes that.

The first commit is just cleanup of the core settings plugin.

### DIFF
--- a/src/main/scala/org/allenai/plugins/CoreSettingsPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/CoreSettingsPlugin.scala
@@ -1,6 +1,7 @@
 package org.allenai.plugins
 
-import com.typesafe.sbt.SbtScalariform
+import com.typesafe.sbt.{ SbtGit, SbtScalariform }
+import com.typesafe.sbt.git.{ DefaultReadableGit, GitReadonlyInterface, ReadableGit }
 import sbt.{
   // Implicit conversion to allow `%` on String.
   toGroupID,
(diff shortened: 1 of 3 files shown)


src/main/scala/org/allenai/plugins/CoreSettingsPlugin.scala
```diff
@@ -1,14 +1,36 @@
 package org.allenai.plugins
 
-import com.typesafe.sbt.SbtScalariform
-import sbt.Keys._
-import sbt._
-
+import com.typesafe.sbt.{ SbtGit, SbtScalariform }
+import com.typesafe.sbt.git.{ DefaultReadableGit, GitReadonlyInterface, ReadableGit }
+import sbt.{
+  // Implicit conversion to allow `%` on String.
```
ColinArenz: It's not clear what this comment applies to.  I assume `toGroupID` is a special implicit converter?

src/main/scala/org/allenai/plugins/CoreSettingsPlugin.scala
```diff
@@ -1,14 +1,36 @@
...
+  Keys,
+  MessageOnlyException,
+  PathFinder,
+  PluginTrigger,
+  Plugins,
+  Project,
+  Setting,
+  Test,
+  ThisBuild
+}
```
ColinArenz: Given how long this list is, it may be best to split it out into a bunch of fully qualified imports?

src/main/scala/org/allenai/plugins/CoreSettingsPlugin.scala
jkinkead: Yes, it's just for `toGroupID`. I expanded the comment to be more explicit.

src/main/scala/org/allenai/plugins/CoreSettingsPlugin.scala
jkinkead: We usually just do it this way. I don't think either is great, but since it's not longer than a default terminal window (25 lines), fully-qualifying seems odd.

jkinkead:
After struggling with this for a while, I realized that the settings being added were due to some old dependencies, which I've updated.

I likely broke the scaladoc plugin - but it never worked properly anyway, and isn't used by anything, so that's fine.

jkinkead:
Oh, and for the record: My previous fix work automatically due to the order in which settings were being applied to the project. In tracking down this issue, I discovered the outdated git plugin version.

PR #274, opened 2017-01-11
Update release.md

aimichal:


docs/release.md
```diff
@@ -19,7 +19,9 @@ After the plugin is installed, release using the following checklist:
 1.  Make sure you are on a branch of the main repository other than `master`.
 **You cannot use a branch of your fork for releases.**  If you do, your tag
 will not make it to the main repository.
-2.  Set the upstream repository (`git push --set-upstream $upstream_name $branch_name`).
+2.  Set the upstream repository:
+  * `git push --set-upstream $upstream_name $branch_name`
```
jkinkead: I'd inline above.

docs/release.md
```diff
@@ -19,7 +19,9 @@ After the plugin is installed, release using the following checklist:
...
+  * Note that the value of `$upstream_name` should be the output of `git remote`.
```
jkinkead: I would rephrase as:
> `$upstream_name` is the name of the git remote pointing to the main repository, which you can see using `git remote -v`.

Note the `-v` flag, which shows the git URL of the remote. Otherwise, you can't tell which remote(s) point to forks, and which point to the main repository.

PR #275, opened 2017-01-11
Added more logging to the DockerBuild plugin. 

ca16:
Added a couple of lines to log a command used.

### DIFF
--- a/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
@@ -765,7 +765,9 @@ $DOCKERFILE_SIGIL
         case (hostPort, containerPort) => Seq("-p", s"$hostPort:$containerPort")
       }
       // Start up the container.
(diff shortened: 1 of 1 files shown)


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -765,7 +765,9 @@ $DOCKERFILE_SIGIL
         case (hostPort, containerPort) => Seq("-p", s"$hostPort:$containerPort")
       }
       // Start up the container.
-      Process(baseCommand ++ portArgs ++ dockerRunFlags.value :+ mainImageName.value).run()
+      val fullCommand = baseCommand ++ portArgs ++ dockerRunFlags.value :+ mainImageName.value
+      logger.info("Running command: " + fullCommand.mkString(" "))
```
jkinkead: I notice that there's a log message on `753` that also gets printed. Maybe also remove that line?

ca16: Removed!

Issue #276, opened 2017-01-25
First run with private repositories fails


ColinArenz:
When building a project which uses our default private repository, we've seen an error on several machines.  This only happens on machines with Docker newly installed.

Command (in a DockerBuildPlugin-enabled project):
```
sbt dockerBuild
```

Result:
```
Sending build context to Docker daemon 9.157 MB

Step 1/5 : FROM allenai-docker-private-docker.bintray.io/oracle-java:8

Get https://allenai-docker-private-docker.bintray.io/v2/oracle-java/manifests/8: unauthorized: Unauthorized

...
```

There appears to be a workaround by pulling an image manually, after which everything works:
```
docker pull allenai-docker-private-docker.bintray.io/oracle-java:8
```

jkinkead:
What does "newly installed" mean?

Does running a `docker build` from the commandline with the same dependency cause an error?

ColinArenz:
Newly installed: Having freshly installed Docker, set up credentials, etc.  Essentially following our setup guide for the first time on a machine.

Docker build: We can try this next time we see the issue, but we haven't tried it yet.

jkinkead:
Does the problem resolve itself on reboot?

jkinkead:
Also, did this occur for anyone other than @splittingfield ? He was using the beta client and might have hit a bug.

We've had several people set up Docker since he did without running into issues.

ColinArenz:
We haven't tried a reboot of the machine.  Restarting Docker doesn't resolve the issue.

These people have hit the issue so far: @splittingfield, @cindy21td, @jayantk

PR #280, opened 2017-02-27
Update the base docker image before building.

ColinArenz:


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -587,6 +587,16 @@ $DOCKERFILE_SIGIL
     // This task requires that the docker dependency stage have been run.
     dockerDependencyStage.value
 
+    // Ensure that the base image is up to date. This also works around an issue wherein the
+    // `docker build` command will fail to authenticate to a repository even when credentials are
+    // valid. The `docker pull` command doesn't exhibit this issue.
+    // TODO: We should invalidate the dependency hash if this image has changed.
```
ColinArenz: @jkinkead I've left a TODO here instead of doing this because I'm not quite sure how to get command output.  The easiest way to do this would be to run `docker images -q <base-image>`, which returns the image ID (or nothing if the image has never been pulled). We'd then run `docker pull ...`, and finally run `docker images ...` again. If the output is different, invalidate the hash.

ColinArenz:
Fixes #276

PR #282, opened 2017-03-01
Export envVars to Dockerfile.

### LINKED ISSUE #281: DockerBuildPlugin should obey envVars
I have code like the following in my web service's `build.sbt`, so I can define the port in one place:
```sbt
containerPort := 8087 // For jetty:start
dockerPorts += containerPort.value // For dockerRun
envVars += ("PORT" -> containerPort.value.toString) // For run
```

Later, in code, I have:
```scala
sys.env.get("PORT").map(_.toInt).getOrElse(8080)
```

And finally, I need a duplicate directive in the Dockerfile to ensure the service binds itself to the correct port when running in a container:
```Dockerfile
ENV PORT 8087
```

I'd very much like `envVars` to get picked up by the docker plugin, similarly to `javaOptions`.  This would allow me to avoid the duplication in our Dockerfile.

jkinkead:
This works with environment variable names that don't use punctuation; double-quotes and other items
may cause errors. Please stick to [A-Za-z_].

Fixes #281 .

### DIFF
--- a/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
@@ -399,6 +399,25 @@ object DockerBuildPlugin extends AutoPlugin {
     val dockerPortsText = dockerPorts.value.mkString(", ")
     val exposeText = dockerPorts.value.map("EXPOSE " + _).mkString("\n")
 
+    // Create any environment variables specified in sbt.
+    // This throws an error if the environment variable name contains a space, since this is not
+    // handled graciously by Docker.
+    val envVarsText = Keys.envVars.value.map {
+      case (key, value) =>
+        if (key.contains(" ")) {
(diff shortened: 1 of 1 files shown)


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -399,6 +399,25 @@ object DockerBuildPlugin extends AutoPlugin {
...
 
+    // Create any environment variables specified in sbt.
+    // This throws an error if the environment variable name contains a space, since this is not
+    // handled graciously by Docker.
+    val envVarsText = Keys.envVars.value.map {
+      case (key, value) =>
+        if (key.contains(" ")) {
+          sys.error(s"""Environment variable name can't be exported to Docker: \"$key\"""")
+        }
+        val quotedValue = value.replaceAllLiterally("\\", "\\\\").replaceAllLiterally("\"", "\\\"")
```
ColinArenz: This hurts to read, but I have no better suggestion. :)

PR #288, opened 2017-04-14
Automatically log in to ECR if required.

jkinkead:
This should ease the transition for people using sbt to build Docker images.

I like that ECR logs people out for security reasons, but I *don't* like the extra work this requires.

### DIFF
--- a/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
+++ b/src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
@@ -47,6 +47,9 @@ object DockerBuildPlugin extends AutoPlugin {
     */
   val DOCKERFILE_SIGIL = "#+" * 50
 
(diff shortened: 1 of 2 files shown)


src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -836,6 +862,10 @@ $DOCKERFILE_SIGIL
 
     val tags = DefaultParsers.spaceDelimited("<tag>").parsed
 
+    if (!skipEcrLogin.value) {
```
ColinArenz: Why not put this check in the login Call?

jkinkead: Because SBT. It would have to be a dynamic task, and at that point you can't pass an image name.

I could've created two tasks, one for the dependency image, and one for the main image tag - but that becomes far less readable and much more annoying if we need to add a new call.

PR #289, opened 2017-09-15
Convert style warnings to errors

cristipp:
How do I build & publish a new version of the plugins?

### DIFF
--- a/src/main/resources/allenai-style-config.xml
+++ b/src/main/resources/allenai-style-config.xml
@@ -22,7 +22,7 @@
   -->
 
   <!-- Example regular expression checker. This is the most flexible checker available. -->
-  <check level="warning" class="org.scalastyle.file.RegexChecker" enabled="false">
+  <check level="error" class="org.scalastyle.file.RegexChecker" enabled="false">
     <customMessage>No println statements</customMessage>
     <parameters>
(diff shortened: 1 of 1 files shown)


cristipp:
FYI @rlebras

schmmd:
@cristipp what problem are you trying to fix exactly?  I worry that this PR followed by a new release of sbt-plugins will not address your problem.

Currently you are using [allenai-sbt-plugins 1.1.1](https://github.com/allenai/euclid/blob/master/project/plugins.sbt).  A new release would mean you would need to upgrade to [1.5.3](https://github.com/allenai/sbt-plugins/blob/master/version.sbt).  1.1.1 was [released in October 2015](https://github.com/allenai/sbt-plugins/releases/tag/v1.1.1) so I would guess that upgrading would be a nightmare.

What is the exact problem you are running into?

cristipp:
Problem: We have just cleaned up hundreds of scalastyle warnings from our build. We'd like to seal the deal and prevent new scalastyle warnings to be added to the codebase.

cristipp:
Oh, by the way, I intended to make this PR against an _euclid_ branch, but apparently I failed to do so. Give me a few more minutes.

cristipp:
Moving over to #290, which merges changes into euclid branch.

PR #290, opened 2017-09-15
Turn scalastyle warnings into errors

cristipp:


cristipp:
FYI @rlebras

schmmd:
Looks great for merging into the Euclid branch ;-P

cristipp:
Does the release process work with branches as well? @schmmd 
https://github.com/allenai/sbt-plugins#releasing

brendan-ai2:
LGTM. One small concern, have you checked that the mechanism for disabling scalariform on a line-by-line basis works at the error level?

schmmd:
@cristipp you will release it to the world (which is OK).  Be sure to include `euclid` in the version.  I.e. `sbt-plugins-v1.1.1_euclid`.  You can edit the version in `version.sbt`.

https://github.com/allenai/sbt-plugins/blob/master/version.sbt

I have not released this in years, and nobody has released it since March--so you may hit problems.  I can help troubleshoot.

cristipp:
@brendan-ai2 No. But I don't know how test it without the entire ceremony of publishing a new release.

schmmd:
@cristipp you can `sbt publishLocal`.  That will install it into your local repository (`~.ivy2/local` I believe) and you can import the dependency from there.

schmmd:
There's also a test suite set up in this project--but I don't know much about that.

brendan-ai2:
More simply, just try violating one of the existing checks marked as `error`. If you can break it and then bypass it, we're good.

cristipp:
Updated version, will try to "publish".

PR #292, opened 2018-06-01
Remove references to bintray.

aimichal:
To be clear, I intend to remove references to Bintray as they pertain to Docker images.

There remain many references to Bintray for maven, which I'm NOT intending to change.

Edit: Added "NOT" :-)

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -54,7 +54,7 @@ ${dockerDependencyExtra.value.mkString("\n")}
 COPY bin bin
 COPY lib lib
 ```
-The base image defaults to `allenai-docker-private-docker.bintray.io/oracle-java:8`, and the workdir defaults to `/stage`. `bin` is copied in order to capture [the start script](#the-start-script), while `lib` contains all of the dependency jars. Finally, this configures the default stop signal for your docker container (the default SIGTERM is ignored by the JVM).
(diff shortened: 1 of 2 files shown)


docs/docker.md
```diff
@@ -179,7 +179,7 @@ for detailed info on sbt mappings.
 
 ### dockerImageBase
 Value type: `String`  
-Default: `allenai-docker-private-docker.bintray.io/oracle-java:8`
+Default: `allenai-sbt-plugins-default-registry.allenai.org/oracle-java:8`
```
schmmd: Is this an actual URL or a placeholder example?

schmmd:
@aimichal why do you intend to change the references to BinTray for Maven?  I think BinTray is a perfectly acceptable place to store Maven artifacts and we don't have alternatives.

aimichal:
> @aimichal why do you intend to change the references to BinTray for Maven? I think BinTray is a perfectly acceptable place to store Maven artifacts and we don't have alternatives.

I'm leaving those alone. Only removing references to Bintray as they pertain to Docker images.

docs/docker.md
aimichal: It's a placeholder. The doc in the DockerBuildPlugin object is updated to be: "Must be overridden to push images.".

I can make this value even more explicitly fake: `allenai-sbt-plugins-FAKE-REGISTRY.allenai.org`

schmmd: I would make it more explicitly fake.

PR #299, opened 2020-06-11
Improve release process

bbstilson:
When updating this, I added cross sbt versions, but didn't set it up correctly to do so. This should fix it. I published locally with v3.0.0, and it seems like it works. I'm not 100% sure it will work when I run it for reals, and I can't test that until I merge.

In `~/.ivy2/local/org.allenai.plugins/allenai-sbt-plugins`
```
.
├── scala_2.10
│   └── sbt_0.13
│       └── 3.0.0
└── scala_2.12
    └── sbt_1.0
        ├── 3.0.0
```

### DIFF
--- a/project/Release.scala
+++ b/project/Release.scala
@@ -9,7 +9,6 @@ object Release {
 
   def releaseSettings = Seq(
     crossSbtVersions := Vector("1.3.10", "0.13.16"),
-    publishMavenStyle := false,
     bintrayRepository := "sbt-plugins",
     bintrayPackage := name.value,
     bintrayOrganization := Some("allenai"),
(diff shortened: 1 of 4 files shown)


project/Release.scala
```diff
@@ -0,0 +1,58 @@
...
+  def releaseProcessSteps = Seq[ReleaseStep](
+    checkSnapshotDependencies,
+    inquireVersions,
+    runClean,
+    releaseStepCommandAndRemaining("^test"),
+    releaseStepCommandAndRemaining("^ scripted"),
+    setReleaseVersion,
+    commitReleaseVersion,
+    tagRelease,
+    releaseStepCommandAndRemaining("^publish"),
```
bbstilson, line 38: These 3 `releaseStepCommandAndRemaining` changes with the `^` are the main changes. Where `+` tells sbt to run the command with all versions of scala specified in the `crossScalaVersions` key -- for example, `+test` will run `test` on all scala versions -- `^` tells sbt to run the command with all version of sbt specified in the `crossSbtVersions` key.

PR #300, opened 2020-06-11
typo and fix ci

bbstilson:


.circleci/config.yml
```diff
@@ -43,4 +43,4 @@ jobs:
       # check formatting
       - run: cat /dev/null | sbt scalafmtCheckAll
       # run tests
-      - run: cat /dev/null | sbt test
+      - run: cat /dev/null | sbt ^test
```
bbstilson, line 46: Needed to test cross versions.

project/Release.scala
```diff
@@ -40,18 +50,4 @@ object Release {
...
-
-  def publishEnabled = Seq(
-    publishArtifact := true, // Enable publish
-    publishMavenStyle := false,
-    // http://www.scala-sbt.org/0.12.2/docs/Detailed-Topics/Artifacts.html
-    publishArtifact in Test := false,
-    // Bintray
-    bintrayOrganization := Some("allenai"),
-    bintrayPackageLabels := Seq("scala"),
-    bintrayRepository := "private",
```
bbstilson, line 52: This key was specified twice, and last writer wins, so the real fix was just deleting this key. I did a bit of re-org to make it less confusing.

PR #301, opened 2020-06-14
delete style plugin

bbstilson:
This PR guts the style plugin.

### Context

I've been updating S2's common plugin/codebases to cross compile for sbt 0.13 and 1.0 and for all scala versions (2.11, 2.12 and 2.13). In order to cross compile this repo, I needed to upgrade the scalariform plugin as the version we were using was not built for sbt 1.0. In going from v1.6.0 to 1.8.3, the default styles changed, which means everything in scholar is changing. That isn't inherently bad, but trying to figure out how to set styling and get them applied has proven sysphean and unreasonably difficult as I need to push changes to the repo, publish to bintray, then download the latest version in my branch in scholar. (I could publishLocal, but it's still a huge pain as its unclear _where_ to even change the styles).

### Reasoning

- Scalariform is needlessly complex, and having dove into the source code, it's unclear how default styles have changed, but they have. You have to build and execute a jar in order for it to work, which is silly.
- There's no need for us to maintain this. I think we should have a common styling and auto-formatting, but I am going to make the case that we use the [scalafmt](https://scalameta.org/scalafmt/). It's significantly more straightforward, documentation is good, and there is an sbt plugin with defined tasks that will be easy to swap out for our custom defined formatting tasks.
- Formatting will be local to the repo. This makes it clear where the formatting rules are coming from and what they are.

### Tasks

- [x] This affects all of S2, so I need to get input from the whole team.
- [ ] Make a pre-commit hook and update the documentation [here](https://github.com/allenai/scholar/blob/master/online/GettingStarted.md#setup-scala-format-pre-commit-hook) on how to make it for new devs.
- [x] Add the [scalafmt plugin](https://scalameta.org/scalafmt/docs/installation.html#sbt) to scholar (and whatever other repos use the custom `formatCheck` task).
- [ ] Update teamcity CI to use `scalafmtCheckAll` instead of `formatCheck`

Am I missing anything?


bbstilson:
1) Maybe. I envision doing a big formatting PR anyways as it's going to be difficult to match the _eexact_ styling we had before. So long as the new formatting rules are drastically different (or worse), there's no harm having a bit of time where PRs aren't getting format checked. This are the format rules I was planning on using: https://github.com/allenai/s2-scala-libraries/blob/master/.scalafmt.conf

2) Yep. I think most people really don't care. The main purpose of a formatting tool is consistency and to keep PR bickering to a minimum.

3) I'm not sure if it's better or worse in that regard.

4) Yeah that makes sense. Any reason each person needs their own pre-commit config? That seems like something that should be in source.

aschokking:
> there's no harm having a bit of time where PRs aren't getting format checked
I think the harm is extra formatting-based noise/churn in PRs in the mean time. What's the harm in waiting for the replacement to be in place before we pull out the current solution?

> Any reason each person needs their own pre-commit config?
It was added as optional to not be too opinionated about which hooks everyone has to start. That said, I'm totally down for landing on an established base of hooks that everyone in the repo has to use and enabling customization on top of it. 

For instance I have a custom hook that looks for the words `DONT COMMIT THIS` and refuses to let me commit which I use for hacks and config changes I don't want to leak into the repo.

bbstilson:
>What's the harm in waiting for the replacement to be in place before we pull out the current solution?

No harm. Let me set it up.

dirkraft:
Less Is More 🕶️ ✌️

PR #303, opened 2021-03-10
update sbt-plugins

bbstilson:
- drop bintray for codeartifact
- drop sbt 0.13 support
- drop including plugins in plugin

This was successfully published to CodeArtifact.


project/Release.scala
```diff
@@ -41,11 +27,11 @@ object Release {
     inquireVersions,
     runClean,
     releaseStepCommandAndRemaining("^test"),
-    releaseStepCommandAndRemaining("^ scripted"),
```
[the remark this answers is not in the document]
bbstilson: I don't think it makes a difference. Just tidying up. Once I get CI running we will know for sure though.

project/Release.scala
```diff
@@ -1,14 +1,12 @@
...
-import bintray.BintrayKeys._
 import sbtrelease.ReleasePlugin.autoImport._
 import sbtrelease.ReleaseStateTransformations._
 
 object Release {
 
   def releaseSettings = Seq(
-    crossSbtVersions := Vector("1.3.10", "0.13.16"),
-    organization := "org.allenai.plugins",
+    organization := "org.allenai",
```
ca16, line 9: why this change?

bbstilson: consistency. It doesn't functionally change anything, but all of our libraries are under the org.allenai organization.

src/main/scala/org/allenai/plugins/CoreRepositories.scala
```diff
@@ -1,53 +0,0 @@
...
-    val ai2PublicReleases = Resolver.bintrayRepo("allenai", "maven")
-
-    // needed for spray-json:
-    val spray = "spray".at("https://repo.spray.io/")
-
-    val typesafeReleases = "Typesafe Releases".at("https://repo.typesafe.com/typesafe/releases/")
-
-    /** Default set of resolvers that will be added via CoreSettings */
-    val defaults = Seq(
-      spray,
```
ca16, line 26: Do we not need these other resolvers anymore?

project/Release.scala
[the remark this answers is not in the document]
bbstilson: If it does make a difference, then it was never doing anything haha

src/main/scala/org/allenai/plugins/CoreRepositories.scala
bbstilson: Spray and typesafe, no. And I couldn't find anything that used jcenter. It's trivial to add back, and if we need them I will be adding them in scholar most likely.

ca16:
> drop sbt 0.13 support

This means:
- Anything still using sbt 0.13 will be able to keep reading from Bintray until Bintray goes away
- Once Bintray goes away, anything still using sbt 0.13 will have to upgrade so it can use the CodeArtifact version which won't support 0.13 (though hopefully this would happen before Bintray goes away)
- Anything not using sbt 0.13 will be able to keep reading from Bintray until Bintray goes away, or switch to CodeArtifact whenever

src/main/scala-sbt-0.13/org/allenai/plugins/Compat.scala
```diff
@@ -1,22 +0,0 @@
-package org.allenai.plugins
```
ca16, line 1: These two Compat files won't be necessary anymore because we will no longer be publishing across sbt versions?

bbstilson: Correct

project/Release.scala
```diff
@@ -41,11 +27,11 @@ object Release {
     inquireVersions,
     runClean,
     releaseStepCommandAndRemaining("^test"),
-    releaseStepCommandAndRemaining("^ scripted"),
+    releaseStepCommandAndRemaining("^scripted"),
     setReleaseVersion,
     commitReleaseVersion,
     tagRelease,
-    releaseStepCommandAndRemaining("^publish"),
+    releaseStepCommandAndRemaining("^codeArtifactPublish"),
```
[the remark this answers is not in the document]
bbstilson: They wouldn't even necessarily need to fully switch, just add the plugin, which adds the resolver.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
```diff
@@ -783,7 +785,7 @@ $DOCKERFILE_SIGIL
 
     // Verify the dockerfile.
     if (verifyDockerfileOnBuild.value) {
-      val isUnchanged = checkDockerfile.value
+      val isUnchanged = checkDockerfile.value: @sbtUnchecked
```
ca16, line 788: so did this used to not need `@sbtUnchecked` but now needs it?

project/Release.scala
```diff
@@ -18,19 +16,7 @@ object Release {
         "git@github.com:allenai/sbt-plugins.git"
       )
     ),
-    // Publish settings.
-    // http://www.scala-sbt.org/0.12.2/docs/Detailed-Topics/Artifacts.html
-    publishArtifact := true, // Enable publish
     publishArtifact in Test := false,
-    publishMavenStyle := false,
```
ca16, line 25: Do you know if the default is false off the top of your head?
Looking at https://www.scala-sbt.org/1.x/docs/Publishing.html#Modifying+the+generated+POM I would _guess_ it is false but not sure at all.

bbstilson: It's set by the codeartifact plugin.

src/main/scala/org/allenai/plugins/DockerBuildPlugin.scala
bbstilson: it was warning before, but this explicitly silences the warning because the publish step was exiting on warnings. That's also why I moved some of the loggers in here around.

project/Release.scala
ca16: So it looks like we're changing the setting right?
https://github.com/bbstilson/sbt-codeartifact/blob/397b52921f6fbfd22e6fc2923c14232dd1878069/sbt-codeartifact/src/main/scala/codeartifact/CodeArtifactPlugin.scala#L49

Is this a CodeArtifact requirement or something else? Not objecting to it, just trying to understand it.

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
```diff
@@ -95,7 +95,7 @@ object VersionInjectorPlugin extends AutoPlugin {
   val gitSha1Task = gitSha1 := (gitCommand("rev-parse", "HEAD").!!).trim
 
   val gitRemotesTask = gitRemotes := {
-    val remotes = gitCommand("remote").lines.toList
+    val remotes = gitCommand("remote").lineStream.toList
```
ca16, line 98: why?

ca16:
>This was successfully published to CodeArtifact.

Is it relatively easy to get a local version of something that depends on sbt-plugins to add that published version and see if anything obvious breaks? Or does CI do something similar?

bbstilson:
>Is it relatively easy to get a local version of something that depends on sbt-plugins to add that published version and see if anything obvious breaks? 

`publishLocal`!

src/main/scala/org/allenai/plugins/VersionInjectorPlugin.scala
bbstilson: deprecation warning haha

bbstilson: I just fix things when I see them

ca16:
> Is it relatively easy to get a local version of something that depends on sbt-plugins to add that published version and see if anything obvious breaks?

> publishLocal!

I think I didn't phrase my question correctly... I was thinking more like rn, to see if this new version of sbt-plugins breaks anything obvious, have you tried/is it easy to try running something in scholar (or some other project that uses sbt-plugins) using this version of sbt-plugins? (whether its the one published to CodeArtifacts or one published by `publishLocal`)

src/sbt-test/sbt-plugins/simple/project/project/plugins.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("io.github.bbstilson" % "sbt-codeartifact" % "0.1.2")
```
ca16, line 1: project/project like you were saying

bbstilson:
@ca16 here's a PR using the changes in this PR: https://github.com/allenai/scholar/pull/27109

ca16 (approved):
Assuming `publishMavenStyle` works out (whether you keep it as true or not), LGTM.

